PR #3, opened 2021-04-14
OCP-MHC and NHC external remediation rbac role

mshitrit:
Tidy + RBAC

PR #5, opened 2021-04-19
External remediation support

n1r1:
This is changing the code to use the external remediation api instead of annotation added by MHC.


peerassistant/health_checker.go
```diff
@@ -31,22 +35,14 @@ func init() {
...
-func isHealthy(machineName string) poisonPillApis.HealthCheckResponse {
-	machine, err := client.Resource(machineRes).Namespace(machineNamespace).Get(context.TODO(), machineName, metav1.GetOptions{})
+func isHealthy(nodeName string) poisonPillApis.HealthCheckResponse {
+	_, err := client.Resource(pprRes).Namespace(pprNamespace).Get(context.TODO(), nodeName, metav1.GetOptions{})
 	if err != nil {
-		//todo examine the error, maybe it's 404 etc.
+		if apiErrors.IsNotFound(err) {
+			return poisonPillApis.Healthy
+		}
 		return poisonPillApis.ApiError
```
rgolangh, line 43: I suggest to wrap this error with the underlying one.

controllers/controller_test.go
```diff
@@ -0,0 +1,184 @@
...
+
+const (
+	nodeName     string = "node1"
+	pprNamespace string = "default"
+)
+
+var _ = Describe("ppr Controller", func() {
+
+	Context("Unhealthy node with api-server access", func() {
+		logf.SetLogger(zap.New(zap.WriteTo(GinkgoWriter), zap.UseDevMode(true)))
```
rgolangh, line 25: This should have been already declared in BeforeSuite

controllers/controller_test.go
```diff
@@ -0,0 +1,184 @@
...
+
+		It("Disable api-server failure simulation", func() {
+			shouldReboot = false
+			apiReaderWrapper.ShouldSimulateFailure = false
```
rgolangh, line 29: you have no expects in this spec.

rgolangh (commented):
I ran a short review. If you prefer to just merge then go a head and I'll continue in codestream

peerassistant/health_checker.go
n1r1: why?
the absence of the CR means the node is healthy

controllers/suite_test.go
```diff
@@ -127,63 +120,19 @@ var _ = BeforeSuite(func(done Done) {
...
-	machineName := "machine1"
-
-	node1 := &v1.Node{}
-	node1.Name = myNodeName
-	node1.Annotations = make(map[string]string)
-	node1.Annotations[machineAnnotationKey] = machineNamespace + "/" + machineName
-	Expect(k8sClient.Create(context.TODO(), node1)).To(Succeed())
-
-	Expect(os.Setenv(nodeNameEnvVar, node1.Name)).To(Succeed())
+	//k8sClient, err = client.New(cfg, client.Options{Scheme: scheme.Scheme})
```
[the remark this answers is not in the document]
n1r1: nah, this should be removed.
thanks!

config/rbac/role.yaml
```diff
@@ -1,104 +1,34 @@
-apiVersion: v1
-kind: ServiceAccount
-metadata:
-  name: manager-role
-  namespace: default
+
 ---
 apiVersion: rbac.authorization.k8s.io/v1
```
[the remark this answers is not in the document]
n1r1: @mshitrit added the rbac stuff for NHC:
https://github.com/medik8s/poison-pill/pull/3

it doesn't involve kubebuilder AFAIU.
If there's a better way to do it I'll be happy to learn

controllers/poisonpillremediation_controller.go
```diff
@@ -137,268 +127,175 @@ func (r *MachineReconciler) Reconcile(ctx context.Context, request ctrl.Request)
...
-		r.Log.Error(err, "failed to retrieve watchdog timeout")
-	} else {
-		r.Log.Info("retrieved watchdog timeout", "timeout", watchdogTimeout)
-		//todo reconcileInterval = watchdogTimeout / 2
+	if !node.Spec.Unschedulable {
+		//the unhealthy node might reboot itself and take new workloads
+		//since we're going to delete the node eventually, we must make sure the node is deleted
+		//when there's no running workload there. Hence we mark it as unschedulable.
+		return r.markNodeAsUnschedulable(node)
 	}
```
mshitrit, line 153: Consider extracting node treatment logic to a method

controllers/poisonpillremediation_controller.go
```diff
@@ -137,268 +127,175 @@ func (r *MachineReconciler) Reconcile(ctx context.Context, request ctrl.Request)
...
-			}
-			return ctrl.Result{}, nil
+	err := r.Client.Status().Update(context.Background(), ppr)
+	if err != nil {
+		if apiErrors.IsConflict(err) {
+			return ctrl.Result{RequeueAfter: 1 * time.Second}, nil
 		}
-
-		r.Log.Error(err, "failed to create node")
+		r.Log.Error(err, "failed to update status with node back up and time to assume node has rebooted")
```
mshitrit, line 204: maybe consider better phrasing.
IIUC : `"failed to update status with node back up and time, assuming node has rebooted"`
Also IIUC this is a valid scenario - so we might want to use something less harsh than an Error

controllers/poisonpillremediation_controller.go
```diff
@@ -80,26 +75,24 @@ var (
...
 	}
 
-	// set myMachineName global variable to the machine this pod runs on
-	if myMachineName == "" {
-		return r.getMachineName()
-	}
-
 	//try to create watchdog if not exists
 	if watchdog == nil {
 		if wdt.IsWatchdogAvailable() {
```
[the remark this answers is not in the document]
n1r1: Good point

controllers/poisonpillremediation_controller.go
```diff
@@ -137,268 +127,175 @@ func (r *MachineReconciler) Reconcile(ctx context.Context, request ctrl.Request)
...
-	return ctrl.Result{RequeueAfter: reconcileInterval}, nil
-}
-
-func (r *MachineReconciler) initWatchdog() error {
-	r.Log.Info("starting watchdog")
-	w, err := wdt.StartWatchdog()
-	watchdog = wdt.Watchdog(w)
-	if err != nil {
-		return err
+	if node.CreationTimestamp.After(ppr.CreationTimestamp.Time) {
```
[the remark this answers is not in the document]
n1r1: good catch.
I believe I need to `return` if there's an err (which is not 404)

peerassistant/health_checker.go
n1r1: re-reading this I realized you pointed the api-error and not the healthy response, so now I understand your comment.

ATM I don't see a real need to propagate the error reason, and this could possibly increase the payload size + make the api a bit more complicated which is something I was trying to avoid for now.

controllers/poisonpillremediation_controller.go
mshitrit: Sounds good, I also understand now what you meant :smile: 
But if you go that way I'd also do the same for `node back up`
i.e. :  `r.Log.Error(err, "failed to update status with node-back-up and time-to-assume-node-has-rebooted")`
Another option is:
`r.Log.Error(err, "failed to update status with 'node back up' and 'time to assume node has rebooted'")`

api/v1alpha1/poisonpillremediation_types.go
```diff
@@ -0,0 +1,75 @@
...
+
+	//NodeBackup is the node object that is going to be deleted as part of the remediation process
+	// +optional
+	// +kubebuilder:validation:EmbeddedResource
+	// +kubebuilder:pruning:PreserveUnknownFields
+	NodeBackup *v1.Node `json:"nodeBackup,omitempty"`
+
+	//TimeAssumedRebooted is the time by then the unhealthy node assumed to be rebooted
+	// +optional
+	TimeAssumedRebooted *metav1.Time `json:"timeAssumedRebooted,omitempty"`
```
beekhof, line 44: What's the default behaviour if no value is supplied?

n1r1: Note that this value is in the `status` section. 
We use this to note the time by which we assume the node already rebooted.
once it's empty, we just update it to `now()`+`node-reboot-timeout`

controllers/poisonpillremediation_controller.go
```diff
@@ -80,26 +75,24 @@ var (
...
 }
 
-// +kubebuilder:rbac:groups=machine.openshift.io.example.com,resources=machines,verbs=get;list;watch;create;update;patch;delete
-// +kubebuilder:rbac:groups=machine.openshift.io.example.com,resources=machines/status,verbs=get;update;patch
-func (r *MachineReconciler) Reconcile(ctx context.Context, request ctrl.Request) (ctrl.Result, error) {
+//+kubebuilder:rbac:groups=poison-pill.medik8s.io,resources=poisonpillremediations,verbs=get;list;watch;create;update;patch;delete
+//+kubebuilder:rbac:groups=poison-pill.medik8s.io,resources=poisonpillremediations/status,verbs=get;update;patch
+//+kubebuilder:rbac:groups=poison-pill.medik8s.io,resources=poisonpillremediations/finalizers,verbs=update
+func (r *PoisonPillRemediationReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Result, error) {
+	_ = r.Log.WithValues("poisonpillremediation", req.NamespacedName)
```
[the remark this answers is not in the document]
slintes: I think you even need use the returned logger, this does not modify an existing one:
`logger := r.Log.WithValues(...)`, and then `logger.Error(...)`

slintes:
general question, maybe I miss something:
- when a PoisonPillRemediation CR is reconciled, the node has a problem already, right?
- when the reconciler can´t talk to the API server, it will ask other nodes
- but the list of other nodes is only filled during reconcilation, so when we already have a problem. So it can´t ask other nodes in case it can´t talk to the api server. Which renders an important part of PP useless....? 🤔

n1r1:
Good questions.

> * when a PoisonPillRemediation CR is reconciled, the node has a problem already, right?

right
> * when the reconciler can´t talk to the API server, it will ask other nodes

correct
> * but the list of other nodes is only filled during reconcilation, so when we already have a problem. So it can´t ask other nodes in case it can´t talk to the api server. Which renders an important part of PP useless....? 

That's a valid point. 
However, we keep the list of nodes in memory.
So we assume that at least once we could fetch the nodes list, we store it in memory, and even if we can't reach the api-server we still have the nodes details.
See this:
https://github.com/medik8s/poison-pill/blob/071a205e2b31b4276cad8c9165f12d8eee615f17/controllers/poisonpillremediation_controller.go#L433-L434

slintes:
> However, we keep the list of nodes in memory.

yes, but the list is never fetched while the node is healthy?

also, thinking one step back: when the node can´t reach the api server, how will will it ever see the remediation CR...? 🤔

n1r1:
> yes, but the list is never fetched while the node is healthy?

why not?
when the node's healthy it would reach this:
https://github.com/medik8s/poison-pill/blob/071a205e2b31b4276cad8c9165f12d8eee615f17/controllers/poisonpillremediation_controller.go#L128

> also, thinking one step back: when the node can´t reach the api server, how will will it ever see the remediation CR...?

it won't.
it will ask its peers if it's healthy or not, and they will try to retrieve the CR for him.
if the CR exists, they will tell him that it's unhealthy.
if the node can't reach the api-server, nor any other peer, he would assume it's unhealthy

n1r1:
I decided to merge this, I will address the open issues on a follow up PR.
thanks everyone for reviewing this

PR #9, opened 2021-04-30
WIP POC: Use separate loops for watchdog and peer healthcheck...

slintes:
... instead of the reconcile function.

Status:
- done: watchdog
- done: peers updates
- done: move error handling to peers, used by peers update and reconcile

Todo: update and extend tests


controllers/poisonpillremediation_controller.go
```diff
@@ -21,7 +21,7 @@ import (
 	"errors"
 	"fmt"
 	poisonPill "github.com/medik8s/poison-pill/api"
-	"github.com/medik8s/poison-pill/utils"
+	"github.com/medik8s/poison-pill/pkg/utils"
```
n1r1, line 24: can you explain the motivation for doing this? (not saying it's wrong just trying to learn)
Also, I think you need to modify the Dockerfile accordingly

slintes: it's good practice to have a `pkg` or `internal` directory, instead of having potentially many top level packages. Since I'm going to introduce at least 1 new package (the loop for updating node list etc) I thought it's a good idea to introduce it now. I was tempted to also move the controllers here, but that will probaly break code scaffolding with operator-sdk.
But since it isn't really related to the topic of this PR I can revert this change if you prefer.

n1r1: > it's good practice to have a `pkg` or `internal` directory, instead of having potentially many top level packages.

can you explain why it's considered a good practice?
I think that Operator-SDK used to generate this `pkg` folder but kubebuilder don't?

> But since it isn't really related to the topic of this PR I can revert this change if you prefer.

No need. I don't have any preference on this subject, so if you think that `pkg` is better I'm all for it.

pkg/watchdog/watchdog.go
```diff
@@ -25,25 +28,62 @@ type watchdogInfo struct {
...
+	}
 
-func StartWatchdog() (*LinuxWatchdog, error) {
 	wdFd, err := openDevice()
 	if err != nil {
 		return nil, fmt.Errorf("failed to open LinuxWatchdog device %s: %v", watchdogDevice, err)
 	}
 
-	wd := &LinuxWatchdog{fd: wdFd,
+	stop := make(chan interface{})
```
n1r1, line 45: why not `chan bool`?

pkg/watchdog/watchdog.go
```diff
@@ -25,25 +28,62 @@ type watchdogInfo struct {
...
+	// feed until stopped
+	ticker := time.NewTicker(*wdTimeout / 3)
+	go func() {
+		for {
+			select {
+			case <-stop:
+				return
+			case <-ticker.C:
+				if err := wd.feed(); err != nil {
+					fmt.Println(fmt.Errorf("failed to feed watchdog! %v", err))
```
n1r1, line 67: does output from `fmt.Println` gets into the container's log?

n1r1 (commented):
Took a look mainly on the intereseting and important stuff and I like it. I think this is a better way to feed the watchdog. thanks!

pkg/watchdog/dummy_watchdog.go
```diff
@@ -0,0 +1,45 @@
...
+		stop: stop,
+	}
+
+	// feed until stopped
+	ticker := time.NewTicker(timeout / 3)
+	go func() {
+		for {
+			select {
+			case <-stop:
+				return
```
n1r1, line 30: I think this means that once a consumer called `Stop()` it would be blocked forever in the second time it calls `Stop()`.
Do you think we should protect `Stop()` with `once`?

n1r1: practically it means that the `Reconcile()` function will be blocked as it call `Stop()` multiple times, and it won't be able to read more events from the queue. Not sure what are the implications but I think we should avoid this.

n1r1: funny but that's exactly what fails the tests.
I wanted to see why they are failing and the reconcile func was blocked. 
I've submitted a fix for the dummy watchdog. if you're happy with that we can do the same to `linuxWatchdog`

n1r1: test passed but failed on docker build:
https://github.com/medik8s/poison-pill/pull/9#discussion_r624925300

controllers/poisonpillremediation_controller.go
slintes: I barely remember seeing a repo without pkg... 🤔 
Looking for an official source I found https://github.com/golang-standards/project-layout
But I also found e.g.  https://github.com/golang-standards/project-layout/issues/10 and https://github.com/golang-standards/project-layout/issues/117
So it does not seem to be such a good practice everyone agrees with than I thought. Thanks for keeping asking, TIL! 🙂

pkg/watchdog/watchdog.go
```diff
@@ -14,9 +13,13 @@ const (
 	watchdogDevice = "/dev/watchdog1"
 )
 
-type LinuxWatchdog struct {
-	fd   int
-	info *watchdogInfo
+var _ Watchdog = &linuxWatchdog{}
```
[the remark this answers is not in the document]
slintes: this enforces that linuxWatchdog implements the Watchdog interface. If it doesn't, you immediately see it in the editor, and it offers to implement missing methods.

pkg/watchdog/watchdog.go
slintes: Oh good catch! But:
Since `bool` does not add any value (we just need to send "something" to the channel) I always use the cheapest (memory wise) thing available in Go for such channels, and that is `struct{}` (not `interface{}`). Example: https://blog.golang.org/pipelines#TOC_6.

pkg/watchdog/watchdog.go
slintes: not sure tbh, I should use a real logger anyway

pkg/watchdog/dummy_watchdog.go
slintes: Good catch! And interesting solution 👍🏼

slintes: I think I would have used a non blocking send like here: https://gobyexample.com/non-blocking-channel-operations
Just for the record
No clue which one is better

slintes: >  failed on docker build

fixed, actually a good reason for `pkg`, no need to add every new package to the Dockerfile anymore 😉

slintes:
Thx for the review @n1r1 👍🏼  All issues addressed. Will go on with the other loop tomorrow.

pkg/watchdog/dummy_watchdog.go
n1r1: > I think I would have used a non blocking send like here: https://gobyexample.com/non-blocking-channel-operations
> Just for the record
> No clue which one is better

Nice, didn't know it's possible.
I'm concerned this might be risky. `Once` blocks until the message was sent (at least once), while the non-blocking-send is best-effort; if I can send a message - send it, otherwise, don't send it. 
The caller for `Stop()` can't really know if the stop succeeded or not. We can pass a return value but it puts a burden on the caller.

When someone calls `Stop()` we want to make sure that the watchdog being stopped, and if `Stop()` returns, the caller should be able to assume that the wathcdog feeding has stopped.

pkg/watchdog/watchdog.go
n1r1: I wouldn't be worried much from memory size, the channel has only one element capacity, so we're talking on 0 bytes for empty struct, 1 byte for `bool` (without padding), 16 bytes for `interface{}`

but it's good to know about that :+1:

controllers/poisonpillremediation_controller.go
peterbourgon: Hello! `pkg/` is neither good practice nor bad practice, it merely keeps your Go code reliably segregated into well-defined locations in the repo. This means (a) you can specifically target Go files only with commands like `go test ./{cmd,pkg,internal}/...`, and (b) you're not left wondering if the top-level folder `hack/` is a collection of scripts or a Go package.

pkg/watchdog/dummy_watchdog.go
peterbourgon: Hello! A better way to do this is 

```go
func (d *dummyWatchdog) Stop() {
    close(d.stop)
}
```

This would prevent the dummyWatchdog from being Started again, but that usually isn't a problem.

An _even better_ way of doing this is to not use Start and Stop methods and an autonomous goroutine at all!

```go
func (d *dummyWatchdog) Run(ctx context.Context) error {
    ticker := time.NewTicker(timeout/3)
    defer ticker.Stop()
    for {
        select {
        case <-ctx.Done():
            return ctx.Err()
        case now := <-ticker.C:
            d.lastFoodTime = now
        }
    }
}
```

Allowing/expecting your users to `go d.Run(...)` and manage the goroutine themselves leaves concurrency to the caller, always a good idea! :)

pkg/watchdog/dummy_watchdog.go
```diff
@@ -0,0 +1,45 @@
...
+}
+
+func StartDummyWatchdog() Watchdog {
+	stop := make(chan interface{})
+	d := &dummyWatchdog{
+		stop: stop,
+	}
+
+	// feed until stopped
+	ticker := time.NewTicker(timeout / 3)
```
peterbourgon, line 25: Oops! You never Stop this ticker, so it's a memory leak!

pkg/watchdog/dummy_watchdog.go
```diff
@@ -0,0 +1,45 @@
...
+	}()
+	return d
+}
+
+func (d *dummyWatchdog) Stop() {
+	d.stop <- true
+}
+
+func (d *dummyWatchdog) LastFoodTime() time.Time {
+	return d.lastFoodTime
```
peterbourgon, line 47: Oops! This variable is written-to by the Started goroutine, and read-from by this method. That's a data race! :(

pkg/watchdog/watchdog.go
```diff
@@ -25,25 +28,62 @@ type watchdogInfo struct {
 	identity        [32]byte
 }
 
-func IsWatchdogAvailable() bool {
-	_, err := os.Stat(watchdogDevice)
-	return !os.IsNotExist(err)
-}
+func StartWatchdog() (Watchdog, error) {
+
+	if _, err := os.Stat(watchdogDevice); err != nil {
```
peterbourgon, line 39: Wow! This is super low level! Is there a reason you need to use Ioctl and syscalls and all that stuff, instead of just using a time.Ticker?

pkg/watchdog/dummy_watchdog.go
n1r1: Hi @peterbourgon 
Thanks for stepping in!


> A better way to do this is
> 
> ```go
> func (d *dummyWatchdog) Stop() {
>     close(d.stop)
> }
> ```

The problem with this approach is that the caller might call `Stop()` multiple times, and this will fail on the second time, as we can't close a closed channel. We can wrap this with `Once` maybe.

btw, we're hiring, if you're interested to work with us :wink:

pkg/watchdog/watchdog.go
n1r1: We're utilizing a linux watchdog here (either `softdog` or hardware watchdog).
Any software based watchdog (like using a time.Ticker) is less reliable:

1. it runs in user space (softdog runs as a kernel module)
2. bugs could occur 
3. it might be affected from system resources starvation

pkg/watchdog/dummy_watchdog.go
slintes: > while the non-blocking-send is best-effort; if I can send a message - send it, otherwise, don't send it.

yes, but the only possible reason that the send fails is that we already stopped teh watchdog, isn't it?

> A better way to do this is

yes, nice, thanks!

> Allowing/expecting your users to go d.Run(...) and manage the goroutine themselves leaves concurrency to the caller, always a good idea! :)

fair point, but since we are the only users of the code, I personally prefer to hide the complexity of concurrency from the caller.

pkg/watchdog/dummy_watchdog.go
slintes: yup, noticed it already myself, thanks!

pkg/watchdog/dummy_watchdog.go
slintes: good catch, thanks

pkg/watchdog/watchdog.go
slintes: It is low level for a reason: This Poison Pill component is about remediation / fencing of nodes with a problem, and we use the watchdog device (it can even be a hardware device of the server) for rebooting the node when we stop feeding it.

peterbourgon: > We're utilizing a linux watchdog here (either softdog or hardware watchdog).
Any software based watchdog (like using a time.Ticker) is less reliable:
> 
> 1. it runs in user space (softdog runs as a kernel module)
> 2. bugs could occur
> 3. it might be affected from system resources starvation

Hmm. Have you validated those assumptions? :)

> This Poison Pill component is about remediation / fencing of nodes with a problem...

Are your nodes just commodity amd64/arm64/etc. Linux servers? Or are they special hardware?

pkg/watchdog/dummy_watchdog.go
peterbourgon: > I personally prefer to hide the complexity of concurrency from the caller.

One of the core Go proverbs is "leave concurrency to the caller" :) But, your call!

slintes:
- added new peers struct with its own update loop
- moved error handling to the peers struct for itself and for the reconcile func

tests will fail, they need to be updated and extended

pkg/peers/peers.go
```diff
@@ -0,0 +1,332 @@
...
+	return nil
+}
+
+func (p *Peers) updatePeers(ctx context.Context, ignoreError bool) {
+	p.mutex.Lock()
+	defer p.mutex.Unlock()
+
+	nodes := &v1.NodeList{}
+	// get some nodes, but not ourself
+	if err := p.List(ctx, nodes, client.Limit(nrOfPeers), client.MatchingLabelsSelector{Selector: p.peerSelector}); err != nil {
```
n1r1, line 108: if you're not using ApiReader you can get a result from the cache even the api-server is down (which is okay unless you count on this to increase the error count.

In addition, unless `ctx` has timeout this will block forever.

pkg/peers/peers.go
```diff
@@ -0,0 +1,332 @@
...
+
+	for _, address := range chosenNodesAddresses {
+		go p.getHealthStatusFromPeer(address, responsesChan)
+	}
+
+	healthyResponses, unhealthyResponses, apiErrorsResponses, _ := p.sumPeersResponses(nodesBatchCount, responsesChan)
+
+	if healthyResponses > 0 {
+		p.log.Info("Peer told me I'm healthy.")
+		p.reset()
```
n1r1, line 193: We should call this whenever we successfully reach the api-server, so 3 errors in a month don't trigger peers query

controllers/poisonpillremediation_controller.go
```diff
@@ -95,47 +80,33 @@ type PoisonPillRemediationReconciler struct {
...
 	//define context with timeout, otherwise this blocks forever
-	ctx, cancelFunc := context.WithTimeout(context.TODO(), apiServerTimeout)
+	ctx, cancelFunc := context.WithTimeout(ctx, apiServerTimeout)
 	defer cancelFunc()
 
 	ppr := &v1alpha1.PoisonPillRemediation{}
-	//we use ApiReader as it doesn't use the cache. Otherwise, the cache returns the object
-	//even though the api server is not available
-	err := r.ApiReader.Get(ctx, req.NamespacedName, ppr)
+	err := r.Get(ctx, req.NamespacedName, ppr)
```
n1r1, line 88: if the api-server is down you'll get a cached response, if exists in cache. Also, this could block forever if api-server is down

slintes:
closing in favor of separating stuff into multiple PRs.
1st one with watchdog refactoring: https://github.com/medik8s/poison-pill/pull/14

controllers/poisonpillremediation_controller.go
```diff
@@ -95,47 +80,33 @@ type PoisonPillRemediationReconciler struct {
...
-	} else {
-		if err := watchdog.Feed(); err != nil {
-			r.logger.Error(err, "failed to feed watchdog")
-			return ctrl.Result{RequeueAfter: 1 * time.Second}, err
-		}
-	}
-
 	//define context with timeout, otherwise this blocks forever
-	ctx, cancelFunc := context.WithTimeout(context.TODO(), apiServerTimeout)
+	ctx, cancelFunc := context.WithTimeout(ctx, apiServerTimeout)
```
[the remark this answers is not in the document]
slintes: deleted (in new PR)

pkg/peers/peers.go
slintes: ah yes, thanks for the heads up!  done (in new PR)

slintes:
PR nr 2: https://github.com/medik8s/poison-pill/pull/15

PR #10, opened 2021-05-02
Run reconcile on periodic intervals

n1r1:
instead of requesting a requeue, add a channel watch which being fed with even on periodic interval.
This has 2 main purposes:
1. Reduces the burden from the developer to remembet to requesta requeue after each operation, which redueces the chance of a bug
2. If reconcile was never triggered since no CR was created, the wd will never be initallized and poison pill won't be able to detect when api-server errors occur.

### DIFF
--- a/controllers/poisonpillremediation_controller.go
+++ b/controllers/poisonpillremediation_controller.go
@@ -30,6 +30,9 @@ import (
 	"net/http"
 	"os"
 	"os/exec"
+	"sigs.k8s.io/controller-runtime/pkg/event"
+	"sigs.k8s.io/controller-runtime/pkg/handler"
+	"sigs.k8s.io/controller-runtime/pkg/source"
 	"strconv"
 	"time"
 
@@ -104,7 +107,7 @@ func (r *PoisonPillRemediationReconciler) Reconcile(ctx context.Context, req ctr
 		if wdt.IsWatchdogAvailable() {
 			if err := r.initWatchdog(); err != nil {
 				r.logger.Error(err, "failed to start watchdog device")
-				return ctrl.Result{RequeueAfter: reconcileInterval}, err
+				return ctrl.Result{}, err
 			}
 		}
 	} else {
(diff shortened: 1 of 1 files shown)


controllers/poisonpillremediation_controller.go
```diff
@@ -104,7 +107,7 @@ func (r *PoisonPillRemediationReconciler) Reconcile(ctx context.Context, req ctr
 		if wdt.IsWatchdogAvailable() {
 			if err := r.initWatchdog(); err != nil {
 				r.logger.Error(err, "failed to start watchdog device")
-				return ctrl.Result{RequeueAfter: reconcileInterval}, err
+				return ctrl.Result{}, err
```
slintes, line 110: this will still trigger a reconcile, because you return an error

slintes (commented):
While this technically will work, I still hesitate to lgtm something like this:

- IMHO this misuses the reconcile function. It should reconcile a poison pill remediation CR, which indicates that there is a problem with a node, and not do some regular work which isn't related to such a CR.
- With the new channel you fake a PPR for the current node. But there is no problem with the current node. So the reconcile function needs to check if this CR is signaling an actual problem with the current or another node, or just for triggering regular tasks. This is confusing IMHO.

I would like to know what others think about it...

controllers/poisonpillremediation_controller.go
n1r1: is this a problem somehow?

n1r1:
> * IMHO this misuses the reconcile function. It should reconcile a poison pill remediation CR, which indicates that there is a problem with a node, and not do some regular work which isn't related to such a CR.
> * With the new channel you fake a PPR for the current node. But there is no problem with the current node. So the reconcile function needs to check if this CR is signaling an actual problem with the current or another node, or just for triggering regular tasks. This is confusing IMHO.

@slintes 
these are good arguments. 
as for the second point, the reconcile function won't find such CR, this is how it knows there's no actual problem.

but basically I agree.
let's see how the other PR you're working on ends up, and if doesn't end up too risky/complicated let's go with it.

controllers/poisonpillremediation_controller.go
slintes: I understood you wanted to not requeue anymore because of the periodic ones... but that would not make sense for the actual PPR CRs. So I take my comment back, sorry

n1r1:
Closing and going with https://github.com/medik8s/poison-pill/pull/9

PR #12, opened 2021-05-04
bugfix - use the first IP address of the node 

n1r1:
using the node index doesn't make any sense. It's a typo done when we moved to ask peers concurrently.

### DIFF
--- a/controllers/poisonpillremediation_controller.go
+++ b/controllers/poisonpillremediation_controller.go
@@ -557,17 +557,14 @@ func popNodes(nodes *v1.NodeList, count int) []string {
 		return addresses
 	}
 
-	//todo this should be random, check if address exists, etc.
-	nodeIndex := 0 //rand.Intn(len(nodes.Items))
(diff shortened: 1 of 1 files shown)


n1r1:
cc @slintes

controllers/poisonpillremediation_controller.go
```diff
@@ -567,7 +567,7 @@ func popNodes(nodes *v1.NodeList, count int) []string {
 	}
 
 	for i := 0; i < nodesCount; i++ {
-		addresses[i] = chosenNode.Status.Addresses[i].Address //todo node might have multiple addresses
+		addresses[i] = chosenNode.Status.Addresses[0].Address //todo node might have multiple addresses
```
slintes, line 570: you also need to update chosenNode, or use nodes.Items[i]. Else all addresses will be from the same node.

slintes (changes requested):

slintes (approved):

PR #13, opened 2021-05-09
Add a new config controller which install the Daemonset

n1r1:
Migrate from poison pill manager to a controller under the same image and repo. This introduces a new `is-manager` cli arg which determines the mode of the executable (should install DS or should run as agent).
The new controller will reconcile a config CR (which currently has 2 unused fields, to be used in the future).
The new controller uses the `render` and `apply` packages as is.

small piggibacks for this PR (separate commits though):

- adding short names for CRs
- small but important fix for the case where the ppr name is the machine name


controllers/poisonpillconfig_controller.go
```diff
@@ -0,0 +1,118 @@
...
+
+	config := &poisonpillv1alpha1.PoisonPillConfig{}
+	if err := r.Client.Get(context.Background(), req.NamespacedName, config); err != nil {
+		if errors.IsNotFound(err) {
+			return ctrl.Result{}, nil
+		}
+		return ctrl.Result{}, err
+	}
+
+	logger.Info("reconciling...")
```
mshitrit, line 63: might want this at the top - right after logger declaration

controllers/poisonpillconfig_controller.go
```diff
@@ -0,0 +1,118 @@
...
+//+kubebuilder:rbac:groups=machine.openshift.io,resources=machines,verbs=get;list;watch;create;update;patch;delete
+//+kubebuilder:rbac:groups=machine.openshift.io,resources=machines/status,verbs=get;update;patch
+
+func (r *PoisonPillConfigReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Result, error) {
+	logger := r.Log.WithValues("poisonpillconfig", req.NamespacedName)
+
+	config := &poisonpillv1alpha1.PoisonPillConfig{}
+	if err := r.Client.Get(context.Background(), req.NamespacedName, config); err != nil {
+		if errors.IsNotFound(err) {
+			return ctrl.Result{}, nil
```
mshitrit, line 58: is having no config a valid scenario ?

controllers/poisonpillconfig_controller.go
```diff
@@ -0,0 +1,118 @@
...
+		}
+		return ctrl.Result{}, err
+	}
+
+	logger.Info("reconciling...")
+	if err := r.syncConfigDaemonSet(config); err != nil {
```
mshitrit, line 64: in this case we'll return `ctrl.Result{}, nil` is that ok ?

controllers/poisonpillconfig_controller.go
```diff
@@ -0,0 +1,118 @@
...
+	return ctrl.NewControllerManagedBy(mgr).
+		For(&poisonpillv1alpha1.PoisonPillConfig{}).
+		Complete(r)
+}
+
+func (r *PoisonPillConfigReconciler) syncConfigDaemonSet(ppc *poisonpillv1alpha1.PoisonPillConfig) error {
+	logger := r.Log.WithName("syncConfigDaemonset")
+	logger.Info("Start to sync config daemonset")
+
+	objs := []*unstructured.Unstructured{}
```
mshitrit, line 82: nit,
I suggest `var objs []*unstructured.Unstructured`
See [here](https://github.com/golang/go/wiki/CodeReviewComments#declaring-empty-slices)

controllers/poisonpillconfig_controller.go
n1r1: actually I used this mainly for debugging, I might just remove it

controllers/poisonpillconfig_controller.go
n1r1: yes, but this is out of scope for this PR

n1r1: I'd expect the controller to remove the DS if no CR exists, but this would be easy when we answer the first open question

controllers/poisonpillconfig_controller.go
n1r1: good catch.
needs to be fixed.

controllers/poisonpillconfig_controller.go
n1r1: I think I will just delete this line. it's not really needed.
I don't want to modify the external packages I used, especially that this is not only a syntax change. They are semantically different.

apply/merge.go
```diff
@@ -0,0 +1,226 @@
...
+
+	mergeAnnotations(current, updated)
+	mergeLabels(current, updated)
+}
+
+// MergeObjectForUpdate prepares a "desired" object to be updated.
+// Some objects, such as Deployments and Services require
+// some semantic-aware updates
+func MergeObjectForUpdate(current, updated *uns.Unstructured) error {
+	if err := mergeDeploymentForUpdate(current, updated); err != nil {
```
slintes, line 29: this is missing DeamonSets at least

slintes (changes requested):
> do we want to reconcile only a specific config name? seems to be a common pattern

Another option is to reconcile only the first CR, and set other to failed in their status... but I have no strong preference...

> which image to put in the POISON_PILL_IMAGE

CI should first build and push the image, and then update the manifest with the relevant image URL. I'd suggest to use git tags for determining the version / image tag.

install/poison-pill-deamonset.yaml
```diff
@@ -0,0 +1,55 @@
...
+  selector:
+    matchLabels:
+      control-plane: controller-manager
+  template:
+    metadata:
+      creationTimestamp: null
+      labels:
+        control-plane: controller-manager
+    spec:
+      serviceAccountName: poison-pill-controller-manager
```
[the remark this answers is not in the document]
n1r1: Yeah, that thought crossed my mind as well.
I will add this to our backlog.

render/render.go
```diff
@@ -0,0 +1,271 @@
+package render
```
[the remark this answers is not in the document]
n1r1: hmm I'm trying to treat that as an external package. 
Any fork we made could make it a more complicated to sync later if there will be any changes. 
I'm wondering if we should make some of the modification you suggested in the source of truth for this code (if exists)

controllers/poisonpillconfig_controller.go
n1r1: Actually the code should set an owner ref on the DS, such that the cluster will automatically delete the DS once the config CR being deleted. I didn't test this though

controllers/poisonpillconfig_controller_test.go
```diff
@@ -0,0 +1,73 @@
...
+			key := types.NamespacedName{
+				Namespace: namespace,
+				Name:      dsName,
+			}
+			Eventually(func() error {
+				return k8sClient.Get(context.Background(), key, ds)
+			}, 10*time.Second, 250*time.Millisecond).Should(BeNil())
+
+			dsContainers := ds.Spec.Template.Spec.Containers
+			Expect(len(dsContainers), 1)
```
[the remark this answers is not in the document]
n1r1: oops, thanks

n1r1:
Created tickets for all unresolved issues.

render/render.go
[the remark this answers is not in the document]
slintes: If you want to treat this as external package, then why not import it instead of copying?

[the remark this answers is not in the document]
n1r1: good question. 
There's no shared package that does that, i.e. a shared utility package that allows you to install manifests.
we could possibly add dependency on a `https://github.com/openshift/cluster-network-operator/tree/master/pkg/render` but it doesn't feel right. 
In addition, CNF chose to copy it, so I did the same.

PR #14, opened 2021-05-10
Refactor watchdog

slintes:
- run feeding in its own loop
- let the manager handler the lifecycle: call start, disarm using context
- deduplicate code by integrating the fake watchdog into the linux watchdog

also moved all packages except controllers (would break code generation) to `/pkg`


pkg/watchdog/watchdog.go
```diff
@@ -0,0 +1,222 @@
...
+	wd.started = true
+	wd.mutex.Unlock()
+
+	feedCtx, cancel := context.WithCancel(context.Background())
+	wd.stop = cancel
+
+	// feed until stopped
+	go wait.NonSlidingUntilWithContext(feedCtx, func(feedCtx context.Context) {
+		if err := wd.feed(); err != nil {
+			wd.log.Error(err, "failed to feed watchdog!")
```
n1r1, line 88: this part is not protected by the mutex. Is the logger thread-safe?

pkg/watchdog/watchdog.go
```diff
@@ -0,0 +1,222 @@
...
+
+	wd.log.Info("watchdog started")
+
+	<-ctx.Done()
+
+	// pod is being stopped, disarm!
+	wd.mutex.Lock()
+	defer wd.mutex.Unlock()
+	if wd.started && !wd.stopped {
+		if err := wd.disarm(); err != nil {
```
n1r1, line 104: there could be a race here.
Looks like the feeding is not protected by the mutex, so it could run in parallel.
So once this writes the disarm character, the feeding might re-arm the watchdog (it could happen befor disarm() closes the fd)

pkg/watchdog/watchdog.go
```diff
@@ -0,0 +1,222 @@
...
+func NewFake(log logr.Logger) (Watchdog, error) {
+	wd := &linuxWatchdog{
+		mutex: sync.Mutex{},
+		log:   log,
+		fake:  true,
+	}
+	return wd, nil
+}
+
+func New(log logr.Logger) (Watchdog, error) {
```
n1r1, line 54: WDYT about making this a singletone?

pkg/watchdog/watchdog.go
```diff
@@ -0,0 +1,222 @@
...
+	fake         bool
+}
+
+type watchdogInfo struct {
+	options         uint32
+	firmwareVersion uint32
+	identity        [32]byte
+}
+
+func NewFake(log logr.Logger) (Watchdog, error) {
```
n1r1, line 45: Not a big thing, but personally I prefer to use dependency injection for tests rather than adding test code to production code.

what's your take on this?

n1r1 (changes requested):

pkg/watchdog/watchdog.go
```diff
@@ -0,0 +1,222 @@
...
+	return wd, nil
+}
+
+func (wd *linuxWatchdog) Start(ctx context.Context) error {
+	wd.mutex.Lock()
+	if wd.started {
+		wd.mutex.Unlock()
+		return errors.New("watchdog was started more than once. This is likely to be caused by being added to a manager multiple times")
+	}
+	wd.start()
```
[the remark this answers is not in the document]
slintes: oh yes, thanks. Question though: should we handle this as "ok, we go on with soft reboot", or a serious problem which stops the pod? So just log the problem and don't mark the watchdog as started? Or return an error? 🤔

pkg/watchdog/watchdog.go
slintes: hm, good question. I never saw a logger being protected at least. And the worst thing which can happen is logs being in wrong order, right? So I think this is fine.

pkg/watchdog/watchdog.go
slintes: good catch, improved the lock

pkg/watchdog/watchdog.go
slintes: I always hesitate to use singletons because they are often referred to as anti pattern and not pattern.
What would be the advantage here?

See e.g. https://www.michaelsafyan.com/tech/design/patterns/singleton

pkg/watchdog/watchdog.go
slintes: The idea was to prevent too much code duplication while keeping same (or at least very similar) behavior for start / stop / feed etc. I was changing code at 2 places all the time. Do have an example for how dependency injection would look like?

slintes:
addressed feedback, rebased, move render / apply to pkg

pkg/watchdog/watchdog.go
n1r1: When a caller calls `disarm()` it expect it to disarm.
When a caller calls `Stop()` it expect the wd to stop and trigger reboot.

If there are multiple callers which started multiple watchdogs, the would interfere one with each other and will break the expectations I mentioned above, as one might feed the watchdog, after the other stopped it.

pkg/watchdog/watchdog.go
n1r1: Worst case can be much worse IMO. e.g. mixed log lines (instead of "hello" and "world" you get "hellworldo") or even corruption.

pkg/watchdog/watchdog.go
n1r1: The previous implementation is kind of dependency injection. 
if we don't want to duplicate some of the code maybe we can embed a linuxWatchdog in the dummyWatchdog, and call only the functions we're interested in, otherwise do the dummy implementation.

pkg/watchdog/watchdog.go
slintes: ok, this is inside the lock now

pkg/watchdog/watchdog.go
slintes: initiating multiple linux watchdogs is prevented now

pkg/watchdog/watchdog.go
slintes: ok, I refactored it:
We still have the `Watchdog interface`, providing the public methods.
Then we have a `watchdogImpl` interface now, which provides the internal, implementation specific methods.
Then there is a `synchronizedWatchdog`, which does all the sync stuff on the public methods, and calls the internal methods. The internal methods are implemented by a linux and a fake watchdog, which are provided as dependency to the synchronizedWatchdog.
Does that make sense? :)

pkg/watchdog/interface.go
```diff
@@ -0,0 +1,130 @@
...
+	swd.stop = cancel
+	// feed until stopped
+	go wait.NonSlidingUntilWithContext(feedCtx, func(feedCtx context.Context) {
+		swd.mutex.Lock()
+		defer swd.mutex.Unlock()
+		// this should not happen because the context is cancelled already.. but just in case
+		if swd.stopped {
+			return
+		}
+		if err := swd.impl.feed(); err != nil {
```
n1r1, line 78: actually this can happen, and this check is very important.
the cancel func doesn't block until the go routine actually stopped

pkg/watchdog/interface.go
```diff
@@ -0,0 +1,130 @@
...
+	return nil
+}
+
+func (swd *synchronizedWatchdog) IsStarted() bool {
+	swd.mutex.Lock()
+	defer swd.mutex.Unlock()
+	return swd.started
+}
+
+func (swd *synchronizedWatchdog) Stop() {
```
n1r1, line 110: It worth documenting that Stop() could actually trigger a reboot and to mention explictly it doesn't disarm the watchdog

pkg/watchdog/interface.go
```diff
@@ -0,0 +1,130 @@
...
+		return
+	}
+	if swd.started {
+		swd.stop()
+		swd.stopped = true
+	}
+}
+
+func (swd *synchronizedWatchdog) GetTimeout() time.Duration {
+	return swd.timeout
```
n1r1, line 123: this should be protected by the mutex

n1r1 (changes requested):
looks good! mostly nits

pkg/watchdog/watchdog.go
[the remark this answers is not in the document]
n1r1: good question. for now, let's just log the problem. 
in the future we should probably add a config value so the user can decide if it wants watchdog or nothing.
if you agree, let's create a ticket in the backlog for this

pkg/watchdog/watchdog.go
n1r1: nicely done!

slintes: thx :)

pkg/watchdog/interface.go
```diff
@@ -0,0 +1,130 @@
...
+type watchdogImpl interface {
+	start() (*time.Duration, error)
+	feed() error
+	disarm() error
+}
+
+var _ Watchdog = &synchronizedWatchdog{}
+
+// synchronizedWatchdog implements the Watchdog interface with synchronized calls of the implementation specific methods
+type synchronizedWatchdog struct {
```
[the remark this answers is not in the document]
slintes: done

pkg/watchdog/interface.go
```diff
@@ -0,0 +1,130 @@
...
+	impl         watchdogImpl
+	timeout      time.Duration
+	started      bool
```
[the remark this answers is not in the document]
slintes: done

pkg/watchdog/interface.go
```diff
@@ -0,0 +1,130 @@
...
+	stop         context.CancelFunc
+	stopped      bool
```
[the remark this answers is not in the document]
slintes: done

pkg/watchdog/interface.go
```diff
@@ -0,0 +1,130 @@
...
+
+func newSynced(log logr.Logger, impl watchdogImpl) *synchronizedWatchdog {
+	return &synchronizedWatchdog{
+		impl: impl,
+		log:  log,
+	}
+}
+
+func (swd *synchronizedWatchdog) Start(ctx context.Context) error {
+	swd.mutex.Lock()
```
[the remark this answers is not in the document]
slintes: I was unsure because it's unlocked and locked again later... but yes, now I agree that doing the deferred unlock here is nicer

pkg/watchdog/interface.go
slintes: done (for all methods on the interface)

pkg/watchdog/interface.go
slintes: 👍🏼

pkg/watchdog/synchronized.go
```diff
@@ -0,0 +1,114 @@
...
+		impl: impl,
+		log:  log,
+	}
+}
+
+func (swd *synchronizedWatchdog) Start(ctx context.Context) error {
+	swd.mutex.Lock()
+	defer swd.mutex.Unlock()
+	if swd.isStarted {
+		return errors.New("watchdog was isStarted more than once. This is likely to be caused by being added to a manager multiple times")
```
n1r1, line 39: s/isStarted/started :)

pkg/watchdog/synchronized.go
```diff
@@ -0,0 +1,114 @@
...
+	}
+	timeout, err := swd.impl.start()
+	if err != nil {
+		// TODO or return the error and fail the pod's start?
+		return nil
+	}
+	swd.timeout = *timeout
+	swd.isStarted = true
+	swd.log.Info("watchdog isStarted")
```
n1r1, line 48: same here

pkg/watchdog/synchronized.go
```diff
@@ -0,0 +1,114 @@
...
+	swd.mutex.Unlock()
+
+	feedCtx, cancel := context.WithCancel(context.Background())
+	swd.stop = cancel
+	// feed until isStopped
```
n1r1, line 53: s/isStopped/stopped?

pkg/watchdog/synchronized.go
slintes: oh, I should better review refactoring in strings :)

pkg/watchdog/synchronized.go
```diff
@@ -0,0 +1,114 @@
...
+	go wait.NonSlidingUntilWithContext(feedCtx, func(feedCtx context.Context) {
+		swd.mutex.Lock()
+		defer swd.mutex.Unlock()
+		// this should not happen because the context is cancelled already.. but just in case
```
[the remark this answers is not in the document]
slintes: so I was curious and did a test: this happens around 15 times in 1 million calls... good enough for a better comment indeed ;)

PR #32, opened 2021-06-08
API check loop

slintes:
follow up for #15 

Instead of using the peer update loop for the detecting the api server connectivity, this PR introduces an new api check loop, which polls the `readyz` endpoint of the api server. This is much cheaper than getting the full node list. So this new check can run with a short interval, while the peer update loop runs less often now.

### DIFF
--- a/pkg/apicheck/check.go
+++ b/pkg/apicheck/check.go
@@ -65,16 +65,26 @@ func (c *ApiConnectivityCheck) Start(ctx context.Context) error {
 	if err != nil {
 		return err
 	}
-	client := cs.RESTClient()
+	restClient := cs.RESTClient()
 
 	go wait.UntilWithContext(ctx, func(ctx context.Context) {
 
 		readerCtx, cancel := context.WithTimeout(ctx, apiServerTimeout)
 		defer cancel()
 
-		result, err := client.Verb(http.MethodGet).RequestURI("/readyz").Do(readerCtx).Raw()
-		if err != nil || string(result) != "ok" {
-			c.log.Error(err, "failed to check api server")
+		result := restClient.Verb(http.MethodGet).RequestURI("/readyz").Do(readerCtx)
+		failure := ""
+		if result.Error() != nil {
+			failure = fmt.Sprintf("api server readyz endpoint error: %v", result.Error())
+		} else {
+			statusCode := 0
+			result.StatusCode(&statusCode)
+			if statusCode != 200 {
+				failure = fmt.Sprintf("api server readyz endpoint status code: %v", statusCode)
+			}
+		}
+		if failure != "" {
+			c.log.Error(fmt.Errorf(failure), "failed to check api server")
 			if isHealthy := c.handleError(); !isHealthy {
 				// we have a problem on this node
 				c.log.Error(err, "we are unhealthy, triggering a reboot")
--- a/pkg/apicheck/check.go
+++ b/pkg/apicheck/check.go
@@ -11,7 +11,9 @@ import (
 	"github.com/go-logr/logr"
 
 	v1 "k8s.io/api/core/v1"
+	"k8s.io/apiextensions-apiserver/pkg/client/clientset/clientset"
 	"k8s.io/apimachinery/pkg/util/wait"
+	"k8s.io/client-go/rest"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 
 	poisonPill "github.com/medik8s/poison-pill/api"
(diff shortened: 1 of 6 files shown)


pkg/apicheck/check.go
```diff
@@ -0,0 +1,260 @@
...
+
+		if unhealthyResponses > 0 {
+			c.log.Info("Peer told me I'm unhealthy!")
+			return false
+		}
+
+		if apiErrorsResponses > 0 {
+			apiErrorsResponsesSum += apiErrorsResponses
+			//todo consider using [m|n]hc.spec.maxUnhealthy instead of 50%
+			if apiErrorsResponsesSum > len(c.peers.GetPeers().Items)/2 { //already reached more than 50% of the nodes and all of them returned api error
```
n1r1, line 157: it would be better to call `GetPeers()` only once, we already do it before this for loop, so we can keep the length of the nodes in a variable declared outside of this for loop scope.
This is better since:
1. GetPeers() is using a lock which could be locked to update the peer so this adds an unperdictable delay here which could impact the time to reboot a node
2. The list of nodes might change during this for loop.

n1r1 (changes requested):
looks good, left some comments

pkg/apicheck/check.go
```diff
@@ -0,0 +1,260 @@
...
+	}
+}
+
+func (c *ApiConnectivityCheck) Start(ctx context.Context) error {
+
+	cs, err := clientset.NewForConfig(c.cfg)
+	if err != nil {
+		return err
+	}
+	client := cs.RESTClient()
```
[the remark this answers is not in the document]
slintes: fixed

pkg/apicheck/check.go
```diff
@@ -0,0 +1,260 @@
...
+
+	go wait.UntilWithContext(ctx, func(ctx context.Context) {
+
+		readerCtx, cancel := context.WithTimeout(ctx, apiServerTimeout)
+		defer cancel()
+
+		result, err := client.Verb(http.MethodGet).RequestURI("/readyz").Do(readerCtx).Raw()
+		if err != nil || string(result) != "ok" {
```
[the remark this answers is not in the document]
slintes: fixed

pkg/apicheck/check.go
```diff
@@ -0,0 +1,260 @@
...
+			c.log.Error(err, "failed to check api server")
```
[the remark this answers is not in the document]
slintes: fixed

pkg/apicheck/check.go
```diff
@@ -0,0 +1,260 @@
...
+func (c *ApiConnectivityCheck) handleError() bool {
+
+	c.errorCount++
+	if c.errorCount < c.maxErrorsThreshold {
+		c.log.Info("Ignoring api-server error, error count below threshold", "current count", c.errorCount, "threshold", c.maxErrorsThreshold)
+		return true
+	}
+
+	c.log.Info("Error count exceeds threshold, trying to ask other nodes if I'm healthy")
+	nodesToAsk := c.peers.GetPeers().Items
```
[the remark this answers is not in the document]
slintes: I changed peers so that the node list is initialized with an empty list

pkg/apicheck/check.go
slintes: 1. The delay is predictable and small, it's at most 5 seconds (`apiServerTimeout` in peers), and it can happen once in 15 minutes (`peerUpdateInterval` in main)
2. Actually I see this as an advantage, it calculates 50% of the current cluster, not of the cluster 15 minutes ago, in case it updated?

pkg/apicheck/check.go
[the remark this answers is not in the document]
n1r1: even better!

pkg/apicheck/check.go
n1r1: 1. Agreed
2. hmm we start with `nodesToAsk` which doesn't get updated. e.g. if you have 10 nodes, `len(nodesToAsk)` is 10, and then the cluster grows to 30 nodes, we'll never get to `len(c.peers.GetPeers().Items)/2` which is 15.

slintes: this is tricky...
what about returning a `lastUpdateTime` in `getPeers`, and if it changes, we restart the complete loop? I think that's the cleanest solution without adding too much complexity...?

slintes: hm, that's also not great, there is a little chance then to run into an endless loop in case it doesn't finish within 15 minutes

n1r1: this is also a problem as it prolongs the remediation time which should be limited with some upper bound.
I don't see a major issue with working with a snapshot of a given time

slintes: yup, using the initial cluster size now

pkg/apicheck/check.go
```diff
@@ -0,0 +1,271 @@
...
+		if failure != "" {
+			c.log.Error(fmt.Errorf(failure), "failed to check api server")
+			if isHealthy := c.handleError(); !isHealthy {
+				// we have a problem on this node
+				c.log.Error(err, "we are unhealthy, triggering a reboot")
+				if err := c.rebooter.Reboot(); err != nil {
+					c.log.Error(err, "failed to trigger reboot")
+				}
+			} else {
+				c.log.Error(err, "peers did not confirm that we are unhealthy, ignoring error")
```
n1r1, line 95: This is not true when we didn't reach the error threshold.

n1r1: I'm not blocking the PR for this message

PR #39, opened 2021-06-16
peer-assistant: support the case of ppr created by MHC and deduce namespace from last ppr

n1r1:
if node loses api-server connectivity, it asks the peers to check if PPR exists in api-server, but in which namespace?
also, should it look for a ppr with node name or machine name?

this PR addresses this need by using the last seen namespace and last seen ppr owner ref (machine/node)
fixes https://bugzilla.redhat.com/show_bug.cgi?id=1972555

Signed-off-by: Nir <niry@redhat.com>

### DIFF
--- a/controllers/poisonpillremediation_controller.go
+++ b/controllers/poisonpillremediation_controller.go
@@ -46,7 +46,18 @@ var (
 		Key:    "node.kubernetes.io/unschedulable",
 		Effect: v1.TaintEffectNoSchedule,
 	}
+
+	lastSeenPprNamespace    string
+	isLastSeenPprWasMachine bool
 )
+
+func GetLastSeenPprNamespace() string {
+	return lastSeenPprNamespace
+}
+
+func IsLastSeenPprWasMachine() bool {
+	return isLastSeenPprWasMachine
+}
 
 // PoisonPillRemediationReconciler reconciles a PoisonPillRemediation object
 type PoisonPillRemediationReconciler struct {
@@ -88,6 +99,8 @@ func (r *PoisonPillRemediationReconciler) Reconcile(ctx context.Context, req ctr
 		r.logger.Error(err, "failed to get PPR")
 		return ctrl.Result{}, err
 	}
+
+	lastSeenPprNamespace = req.Namespace
 
 	node, err := r.getNodeFromPpr(ppr)
 	if err != nil {
@@ -223,6 +236,7 @@ func (r *PoisonPillRemediationReconciler) getNodeFromPpr(ppr *v1alpha1.PoisonPil
 
 	for _, ownerRef := range ppr.OwnerReferences {
 		if ownerRef.Kind == "Machine" {
+			isLastSeenPprWasMachine = true
 			return r.getNodeFromMachine(ownerRef, ppr.Namespace)
 		}
 	}
--- a/pkg/peerassistant/health_checker.go
+++ b/pkg/peerassistant/health_checker.go
@@ -4,22 +4,29 @@ import (
 	"context"
 	poisonPillApis "github.com/medik8s/poison-pill/api"
 	"github.com/medik8s/poison-pill/api/v1alpha1"
+	"github.com/medik8s/poison-pill/controllers"
+	corev1 "k8s.io/api/core/v1"
 	apiErrors "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/runtime/schema"
 	"k8s.io/client-go/dynamic"
 	"k8s.io/client-go/rest"
+	"k8s.io/client-go/tools/cache"
 	"sigs.k8s.io/controller-runtime/pkg/log/zap"
 )
 
-const (
-	pprNamespace = "medik8s"
-)
+const machineAnnotation = "machine.openshift.io/machine" //todo this is openshift specific
 
-var client dynamic.Interface
-var pprRes = schema.GroupVersionResource{Group: v1alpha1.GroupVersion.Group,
-	Version:  v1alpha1.GroupVersion.Version,
-	Resource: "poisonpillremediations"}
+var (
+	client dynamic.Interface
+	pprRes = schema.GroupVersionResource{Group: v1alpha1.GroupVersion.Group,
+		Version:  v1alpha1.GroupVersion.Version,
+		Resource: "poisonpillremediations"}
+	nodeRes = schema.GroupVersionResource{Group: corev1.SchemeGroupVersion.Group,
+		Version:  corev1.SchemeGroupVersion.Version,
+		Resource: "nodes"}
+	logger = zap.New().WithName("health-checker")
+)
 
 func init() {
 	// creates the in-cluster config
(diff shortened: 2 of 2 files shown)


pkg/peerassistant/health_checker.go
```diff
@@ -36,18 +43,57 @@ func init() {
...
+	isMachine := controllers.IsLastSeenPprWasMachine()
+
+	if isMachine {
+		return isHealthyMachine(nodeName, namespace)
+	} else {
+		return isHealthyNode(nodeName, namespace)
+	}
+}
+
+func isHealthyNode(nodeName string, namespace string) poisonPillApis.HealthCheckResponse {
```
mshitrit, line 58: looks like either this method or isHealthyByPpr is redundant

pkg/peerassistant/health_checker.go
```diff
@@ -4,22 +4,29 @@ import (
...
 	"k8s.io/client-go/dynamic"
 	"k8s.io/client-go/rest"
+	"k8s.io/client-go/tools/cache"
 	"sigs.k8s.io/controller-runtime/pkg/log/zap"
 )
 
-const (
-	pprNamespace = "medik8s"
-)
+const machineAnnotation = "machine.openshift.io/machine" //todo this is openshift specific
```
mshitrit, line 18: Since we manage this key on the node annotations, can we use something generic which will fit both cases ?

pkg/peerassistant/health_checker.go
n1r1: semantically it's not redundant

pkg/peerassistant/health_checker.go
n1r1: I don't follow.
the owner of this annotation is MAO

pkg/peerassistant/health_checker.go
mshitrit: Maybe I'm missing something, but the way I see it you have a 1 liner method that all it does is calling another method with the same signature.

let's say I have:
`foo(x int){ //doSomething...}`
and
`goo(x int){ foo(x)}`

I believe that would make `goo` redundant because anywhere it is used I can replace it with `foo` 
if you prefer `goo` semantically delete it and rename `foo` to be `goo`

pkg/peerassistant/health_checker.go
mshitrit: My mistake - didn't realize that

controllers/poisonpillremediation_controller.go
```diff
@@ -46,8 +46,19 @@ var (
 		Key:    "node.kubernetes.io/unschedulable",
 		Effect: v1.TaintEffectNoSchedule,
 	}
+
+	lastSeenPprNamespace    string
```
slintes, line 50: Maybe it's better to make these fields and the getter funcs properties of the reconciler type. Global / package level vars are an anti pattern most times

controllers/poisonpillremediation_controller.go
```diff
@@ -46,8 +46,19 @@ var (
...
+	isLastSeenPprWasMachine bool
```
slintes, line 51: is... was...? Isn´t one boolean indicator enough? This is not improving readability IMHO...

pkg/peerassistant/health_checker.go
```diff
@@ -36,18 +43,57 @@ func init() {
...
 
 func isHealthy(nodeName string) poisonPillApis.HealthCheckResponse {
-	log := zap.New().WithName("health-checker")
-	log.Info("checking health for", "node", nodeName)
-	_, err := client.Resource(pprRes).Namespace(pprNamespace).Get(context.TODO(), nodeName, metav1.GetOptions{})
+	logger.Info("checking health for", "node", nodeName)
+
+	namespace := controllers.GetLastSeenPprNamespace()
+	isMachine := controllers.IsLastSeenPprWasMachine()
+
```
slintes, line 50: Both namespace and isMachine can have their default value when there wasn't a PPR yet?
And even when there was a PPR, how do you know it's namespace and target kind are the same as for this health request? (I guess you can run MHC and NHC in parallel?)

slintes (commented):
it's already merged, but still...

pkg/peerassistant/health_checker.go
mshitrit: Guess we'll have to agree to disagree on that one.
IMHO having two method in the same class with different names that does exactly  the same is just a bad code, no matter how you spin it.

pkg/peerassistant/health_checker.go
slintes: (for the namespace the empty default is even ok, you will look into all namespaces)

pkg/peerassistant/health_checker.go
n1r1: there's no class or methods here.
can you explain why do you think it's a bad code?
is there any reference for that claim?

controllers/poisonpillremediation_controller.go
n1r1: lol.
does `wasLastPprMachine` sounds better?

pkg/peerassistant/health_checker.go
n1r1: I think we discussed this in Slack.
is there still open issue here?

controllers/poisonpillremediation_controller.go
n1r1: makes sense

pkg/peerassistant/health_checker.go
mshitrit: >there's no class or methods here.

It's purely semantic, replace `method` with `func` and `class` in this case would be the scope in which those two `func` exist.

>can you explain why do you think it's a bad code?

It is code duplication - by this logic instead of finding a proper name to a method we can generate multiple method each with a different name that calls the original.

>is there any reference for that claim?

References:
https://rules.sonarsource.com/java/tag/confusing/RSPEC-4144
https://refactoring.guru/smells/duplicate-code

n1r1: The references don't really support your argument, but anyway, let's do it that way - show me how you would implement this otherwise and I'll tell why it's wrong, semantically.

also, you can find other go packages that does similar things, mostly as a syntactic sugar (not always I agree with that, but anyway), see this from ginkgo:
https://github.com/onsi/ginkgo/blob/master/ginkgo_dsl.go#L421-L428

`Specify` and `It` have exactly the same implementation

controllers/poisonpillremediation_controller.go
slintes: I think so ;)

pkg/peerassistant/health_checker.go
slintes: IIRC we discussed that 
> for the namespace the empty default is even ok, you will look into all namespaces

doesn't work for get requests, so we still have to deal with the case that there wasn't a PPR yet and the vars have their default values?

pkg/peerassistant/health_checker.go
slintes: I think it's fine as it is. There is `isHealthyNode` and `isHealthyMachine`, both call `isHealthyByPpr`. Just for the node case you don't need to do anything. I only would have put `isHealthyByPpr` below both others, to make this a bit clearer.
This way the `if` is nicer:
```
if isMachine {
	return isHealthyMachine(nodeName, namespace)
} else {
	return isHealthyNode(nodeName, namespace)
}
```
instead of
```
if isMachine {
	return isHealthyMachine(nodeName, namespace)
} else {
	return isHealthyByPpr(nodeName, namespace)
}
```
(btw, you can skip the `else` because you return in the `if`)

pkg/peerassistant/health_checker.go
slintes: and this doesn't support running MHC and NHC in parallel, which at least in theory can ha[pen, right? 🤔

pkg/peerassistant/health_checker.go
mshitrit: My suggestion was deleting `isHealthyNode`  & refactor `isHealthyMachine` to just the part that retrieves the machine to  `getMachineName`

which would look something like this:

 
```
if isMachine {
        machineName := getMachineName(nodeName, namespace)
	return isHealthyByPpr(machineName, namespace)
} else {
	return isHealthyByPpr(nodeName, namespace)
}
```


Couldn't reach agreement though :disappointed:

slintes: that's nice as well 👍🏼 
but I wouldn't have a strong preference

PR #45, opened 2021-06-27
Protect last seen ppr namespace and is-machine with mutex

n1r1:
there are two go routines involved here - one for the ppr reconciler and the other for the rest server.
Since they weren't protected by a mutex we had a data race.
This was introduced in 7bd2174.
In addition, made some cosmetics changes in e386c4c per @slintes comments on https://github.com/medik8s/poison-pill/pull/39 and extracted agent-init logic and manager-init logic to their own functions

### DIFF
--- a/controllers/poisonpillremediation_controller.go
+++ b/controllers/poisonpillremediation_controller.go
@@ -19,6 +19,7 @@ package controllers
 import (
 	"context"
 	"errors"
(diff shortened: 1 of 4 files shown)


n1r1:
@slintes I'm merging this since I want it to be included in the release.
feel free to add your comments after the merge

PR #65, opened 2021-07-07
Avoid remediation loops by not restoring node's status.

n1r1:
The node's status should be set by the node itself. Otherwise, we restore the node conditions that created the unhealthy state which might lead to remediation loop


slintes:
/lgtm
/test e2e-test

(the latter should have been trigger automatically, but maybe it needs some time)

PR #66, opened 2021-07-07
e2e: check all nodes in the API server issue test case

slintes:
untested, let CI do its work :)


slintes:
/hold

wait for e2e... which can't be triggered yet? 🤔

PR #67, opened 2021-07-08
Prevent potential peer list mutation

rgolangh:
Use a copy of a peer list instead of a reference to prevent mutation
of that list outside of the inner lock. Without this the lock
taked inside any guarded block is just useless because the list
reference just leaks out.

Signed-off-by: Roy Golan <rgolan@redhat.com>

### DIFF
--- a/pkg/peers/peers.go
+++ b/pkg/peers/peers.go
@@ -102,9 +102,9 @@ func (p *Peers) updatePeers(ctx context.Context) {
 	p.peerList = nodes
 }
 
-func (p *Peers) GetPeers() v1.NodeList {
+func (p *Peers) GetPeers() *v1.NodeList {
 	p.mutex.Lock()
 	defer p.mutex.Unlock()
 
-	return p.peerList
+	return p.peerList.DeepCopy()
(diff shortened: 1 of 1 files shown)


rgolangh:
@n1r1 PTAL

n1r1:
@slintes what's your take on this?

n1r1:
in [this line](https://github.com/medik8s/poison-pill/blob/5aa30162c5139b80903e7d3e9d7cec0a6f02dd69/pkg/apicheck/check.go#L116) we have:
`nodesToAsk := c.config.Peers.GetPeers().Items`

and later we modify `nodesToAsk`, will it affect the peers in `peers.go`?
do we have a data race here when `peers.go` is writing to `peers` while it's being used by `check.go`?

slintes:
I think this is a good improvement to make code safer 👍🏼 
However, it isn't an issue in `check.go` today, because it uses the `Items` slice, and it doesn't matter if you get this from a pointer or value "parent", it is always a copy.
Maybe GetPeers() should even return the items directly...?

PR #72, opened 2021-07-09
Reduce flakiness of 2e test

slintes:
- make `killApiConnection` async, the execution time of the synced version isn't reliable
- let `checkNoNodeRecreate` consistently do the check for some minutes instead

### DIFF
--- a/e2e/poison_pill_test.go
+++ b/e2e/poison_pill_test.go
@@ -117,8 +117,8 @@ var _ = Describe("Poison Pill E2E", func() {
 			})
 
 			It("should not reboot and not re-create node", func() {
-				// order matters
-				// - because the 2nd check has a small timeout only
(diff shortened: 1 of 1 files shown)


slintes:
/hold

run e2e a few times to see if this helps

slintes:
/retest

no clue if this makes things better... but I think it also doesn't make things worse

n1r1:
yeah, I think https://github.com/medik8s/poison-pill/pull/72 makes a real improvement.
the last failure is the first to fail on logs

n1r1:
I believe that https://github.com/medik8s/poison-pill/pull/107 has done similar thing but with different approach - blocking until context is expected to expire. e2e test seems to be much more stable now (5 out of 5).
I'm closing this, let me know if you have any objection

PR #73, opened 2021-07-11
WIP: Use ubi-minimal, smaller attack surface, smaller image

rgolangh:
Signed-off-by: Roy Golan <rgolan@redhat.com>

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -19,7 +19,10 @@ COPY install/ install/
 # Build
 RUN CGO_ENABLED=0 GOOS=linux GOARCH=amd64 GO111MODULE=on go build -a -o manager main.go
 
-FROM registry.access.redhat.com/ubi8/ubi:latest
+FROM registry.access.redhat.com/ubi8/ubi-minimal:latest
+# poison pill dependencies
+# util-linux used for /usr/bin/nsenter, used for soft reboot with systemctl
+RUN microdnf install util-linux
 
 WORKDIR /
 COPY --from=builder /workspace/install/ install/


n1r1:
micro didn't contain everything that poison pill needs. I think it was missing `nsenter`

rgolangh:
/hold 
missing nsenter binary in ubi-micro

Dockerfile
```diff
@@ -19,7 +19,10 @@ COPY install/ install/
 # Build
 RUN CGO_ENABLED=0 GOOS=linux GOARCH=amd64 GO111MODULE=on go build -a -o manager main.go
 
-FROM registry.access.redhat.com/ubi8/ubi:latest
+FROM registry.access.redhat.com/ubi8/ubi-minimal:latest
+# poison pill dependencies
+# util-linux used for /usr/bin/nsenter, used for soft reboot with systemctl
+RUN microdnf install util-linux
```
n1r1, line 25: maybe we should use `-y`?

rgolangh: microdnf doesn't have that. It just do what you asked :wink:

n1r1:
@rgolangh what's the status of this?
can we close it?

rgolangh:
ubi-micro is shipped without package management, and I missed that.
To install any dependency, like `util-linux` for `nsenter` the documentation recommends to mount the container in some build machine and run `yum install --installdir` that points at the container image root. More details here  https://access.redhat.com/documentation/en-us/red_hat_enterprise_linux/8/html-single/building_running_and_managing_containers/index#con_understanding-the-ubi-micro-images_assembly_types-of-container-images

This change will require we build the container image using buildah, and not directly using the Dockerfile. 
Since this is too disruptive change I'm closing this.

PR #80, opened 2021-07-15
auto create poison pill remediation template on boot

n1r1:
to make things easier for users

### DIFF
--- a/api/v1alpha1/poisonpillconfig_types.go
+++ b/api/v1alpha1/poisonpillconfig_types.go
@@ -25,6 +25,7 @@ import (
 
 const (
 	configCRName                          = "poison-pill-config"
+	templateCRName                        = "poison-pill-default-template"
 	defaultWatchdogPath                   = "/dev/watchdog1"
 	defaultSafetToAssumeNodeRebootTimeout = 180
 )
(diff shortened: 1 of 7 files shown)


api/v1alpha1/poisonpillconfig_types.go
```diff
@@ -88,3 +89,10 @@ func NewDefaultPoisonPillConfig() PoisonPillConfig {
 		},
 	}
 }
+
+func NewDefaultRemediationTemplate() PoisonPillRemediationTemplate {
+	return PoisonPillRemediationTemplate{
+		ObjectMeta: metav1.ObjectMeta{Name: templateCRName},
+		Spec:       PoisonPillRemediationTemplateSpec{Template: PoisonPillRemediationTemplateResource{Spec: PoisonPillRemediationSpec{}}},
```
rgolangh, line 96: we should add the manager controller as the owner so a remove will clean it up, and a new potentially updated default will be created

n1r1: We can always update it, especially since there's no existing field in the schema.
and I don't want to add complication seconds before ga

bundle/manifests/poison-pill.clusterserviceversion.yaml
```diff
@@ -244,12 +270,11 @@ spec:
                     fieldRef:
                       fieldPath: spec.nodeName
                 - name: POISON_PILL_IMAGE
-                  value: quay.io/medik8s/poison-pill-operator:0.1.1
```
slintes, line 247: heads up, this is removed again!

slintes (changes requested):

bundle/manifests/poison-pill.clusterserviceversion.yaml
n1r1: omg

n1r1: fixed both this specific case, and in general this should not happen again after 7b8282f

PR #81, opened 2021-07-19
Adding an if clause for possible conflict error for r.client .

razo7:
My first PR :)


n1r1:
Thanks @razo7 

/approve

This makes me think that maybe using `patch` would be a better approach. Not sure.
anyway, not in this PR.

could you please just add a comment on the first conflict check that explains why we ignore it?
e.g. 

> conflicts are expected since all poison pill deamonset pods are competing on the same requests

PR #84, opened 2021-07-28
Reconcile only the default config CR

n1r1:
We expect only one poison pill config CR, but we don't enforce this.
This commit ignores ppc that are not in the namespace of the operator or not in the default config name, which practically means it reconciles only one CR.

### DIFF
--- a/controllers/poisonpillconfig_controller_test.go
+++ b/controllers/poisonpillconfig_controller_test.go
@@ -18,12 +18,11 @@ import (
 )
 
 var _ = Describe("ppc controller Test", func() {
+	dsName := "poison-pill-ds"
 
 	Context("DS installation", func() {
 		dummyPoisonPillImage := "poison-pill-image"
 		os.Setenv("POISON_PILL_IMAGE", dummyPoisonPillImage)
-
-		dsName := "poison-pill-ds"
 
 		nsToCreate := &corev1.Namespace{
 			ObjectMeta: metav1.ObjectMeta{
@@ -110,6 +109,40 @@ var _ = Describe("ppc controller Test", func() {
 
 			Expect(createdConfig.Spec.WatchdogFilePath).To(Equal("/dev/watchdog1"))
 			Expect(createdConfig.Spec.SafeTimeToAssumeNodeRebootedSeconds).To(Equal(180))
+		})
+	})
+
+	Context("Wrong poison pill config CR", func() {
+		var dsResourceVersion string
+		ds := &appsv1.DaemonSet{}
+		key := types.NamespacedName{
+			Namespace: namespace,
+			Name:      dsName,
+		}
+
+		It("get DS resource version", func() {
+			Expect(k8sClient.Get(context.Background(), key, ds)).To(Succeed())
+			dsResourceVersion = ds.ResourceVersion
+		})
+
+		config := &poisonpillv1alpha1.PoisonPillConfig{}
+		config.Kind = "PoisonPillConfig"
+		config.APIVersion = "poison-pill.medik8s.io/v1alpha1"
+		config.Name = "not-the-expected-name"
+		config.Spec.WatchdogFilePath = "foo"
+		config.Spec.SafeTimeToAssumeNodeRebootedSeconds = 9999
+		config.Namespace = namespace
+
+		It("Create config CR", func() {
+			Expect(k8sClient).To(Not(BeNil()))
+			Expect(k8sClient.Create(context.Background(), config)).To(Succeed())
+		})
+
+		It("Daemonset should not be changed", func() {
+			Consistently(func() string {
+				Expect(k8sClient.Get(context.Background(), key, ds)).To(Succeed())
+				return ds.ResourceVersion
+			}, 20*time.Second, 1*time.Second).Should(Equal(dsResourceVersion))
 		})
 	})
 
--- a/controllers/poisonpillconfig_controller_test.go
+++ b/controllers/poisonpillconfig_controller_test.go
@@ -114,27 +114,28 @@ var _ = Describe("ppc controller Test", func() {
 
 	Context("Wrong poison pill config CR", func() {
 		var dsResourceVersion string
-		ds := &appsv1.DaemonSet{}
-		key := types.NamespacedName{
-			Namespace: namespace,
-			Name:      dsName,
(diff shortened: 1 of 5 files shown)


slintes:
/lgtm

What about adding a test for this? Feel free to remove the hold though if you think it's not needed
/hold

n1r1:
Good idea. 
I will add a test

controllers/poisonpillconfig_controller_test.go
```diff
@@ -113,6 +112,40 @@ var _ = Describe("ppc controller Test", func() {
...
 
+	Context("Wrong poison pill config CR", func() {
+		var dsResourceVersion string
+		ds := &appsv1.DaemonSet{}
+		key := types.NamespacedName{
+			Namespace: namespace,
+			Name:      dsName,
+		}
+
+		It("get DS resource version", func() {
```
slintes, line 123: I think you are using `It` not quite as it is supposed to. Won't block on it, as I see the other tests use it in the same way, but for the record: one `It` refers to one test. But getting the current DS resource version isn't what you want to test (neither is creating the new config below), it is just preparation. The only thing you actually want to test is that the DS isn't updated. So you need 1 `It` only. In this case multiple `It`s work, but as soon as you also use `BeforeEach` etc. it get's interesting, because they are executed before and after every single `It`.
So IMHO a better way to structure this test could be:
```
Context("Wrong poison pill config CR", func() {
  It("Should not modify the DS", func() {
    By("Getting DS resource version")
    ...
    By("Creating config CR")
    ...
    By("Ensuring DS resource version doesn't change")
    ...
  })
}

File: controllers/poisonpillconfig_controller_test.go [hunk 2]
```diff
@@ -113,6 +112,40 @@ var _ = Describe("ppc controller Test", func() {
...
+		config := &poisonpillv1alpha1.PoisonPillConfig{}
+		config.Kind = "PoisonPillConfig"
+		config.APIVersion = "poison-pill.medik8s.io/v1alpha1"
+		config.Name = "not-the-expected-name"
+		config.Spec.WatchdogFilePath = "foo"
+		config.Spec.SafeTimeToAssumeNodeRebootedSeconds = 9999
+		config.Namespace = namespace
+
+		It("Create config CR", func() {
+			Expect(k8sClient).To(Not(BeNil()))
```
slintes, line 137: this check is redundant, it is in `BeforeSuite()` already (and you used the client above already)

controllers/poisonpillconfig_controller_test.go
```diff
@@ -113,6 +112,40 @@ var _ = Describe("ppc controller Test", func() {
...
+			Expect(k8sClient.Create(context.Background(), config)).To(Succeed())
+		})
+
+		It("Daemonset should not be changed", func() {
+			Consistently(func() string {
+				Expect(k8sClient.Get(context.Background(), key, ds)).To(Succeed())
```
slintes, line 143: hint: be careful with `Expect` within `Eventually`/`Consistently`: it will break the loop and fail the test, so it can potentially introduce flakes, when 1 of the 20 API calls fails for some reason.

slintes (commented):
see comments inline... nothing blocking, but maybe you want to address some, so I'll add the `hold` again

/lgtm
/hold

controllers/poisonpillconfig_controller_test.go
n1r1: The only reason I used `It` because `k8client` works only within It. outside of it, it's just a `nil`
any idea why?

controllers/poisonpillconfig_controller_test.go
n1r1: How would you do that differently?

controllers/poisonpillconfig_controller_test.go
slintes: yes, IIRC: the code in Describe / Context is executed immediately when starting the tests, but the code in BeforeSuite / BeforeEach / It isn't, those "hooks" are only registered for being executed for each test afterwards. So k8sclient in Describe / Context is nil because BeforeSuite didn't run yet.
Rule of thumb IMHO: put everything into BeforeSuite / BeforeEach / It, and use Describe / Context just for structuring the tests without any code inside.

PS: same applies to After* funcs of course

controllers/poisonpillconfig_controller_test.go
slintes: hm, I wanted to look up how to this in the best way (e.g. returning 2 values, which actually works, the 2nd is expected to be nil or zero value, so it can be an error)... but seems I was wrong with my statement at all 🤔 I was sure that I run into an issue with Expect inside Eventually in the past... sorry for the noise

`The function that you pass to Eventually can have more than one return value. In that case, Eventually passes the first return value to the matcher and asserts that all other return values are nil or zero-valued`

`Eventually allows you to make assertions in the passed-in function. The function is assumed to have failed and will be retried if any assertion in the function fails`

`As with Eventually, if you pass Consistently a function that returns more than one value, it will pass the first value to the matcher and assert that all other values are nil or zero-valued. Like Eventually, Consistently also supports functions that make assertions - and will require that all assertions pass consistently. `

https://onsi.github.io/gomega/#making-asynchronous-assertions

n1r1: Thanks. I appreciate your comments

controllers/poisonpillconfig_controller_test.go
n1r1: I wasn't aware of that important stuff.
thanks!

controllers/poisonpillconfig_controller_test.go
n1r1: removed (at least the one I added in this PR)

controllers/poisonpillconfig_controller_test.go
n1r1: I changed the test I added in this PR and later will change all tests to match this.
does it look good to you now?

slintes: yes, looks good 👍🏼 And I actually like using BeforeEach even though there is one test only, make the separation of preparation and actual test very clear 👍🏼

n1r1:
thanks Marc
/unhold

PR #85, opened 2021-07-28
Reduce memory footprint

n1r1:
Before this commit we kept all the node object for all nodes, while all we need is the IP address.

### DIFF
--- a/pkg/apicheck/check.go
+++ b/pkg/apicheck/check.go
@@ -113,7 +113,7 @@ func (c *ApiConnectivityCheck) handleError() bool {
 	}
 
 	c.config.Log.Info("Error count exceeds threshold, trying to ask other nodes if I'm healthy")
-	nodesToAsk := c.config.Peers.GetPeers().Items
+	nodesToAsk := c.config.Peers.GetPeersAddresses()
 	if nodesToAsk == nil || len(nodesToAsk) == 0 {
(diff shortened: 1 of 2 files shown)


pkg/apicheck/check.go
```diff
@@ -191,12 +191,12 @@ func (c *ApiConnectivityCheck) popNodes(nodes *[]v1.Node, count int) []string {
 	//todo maybe we should pick nodes randomly rather than relying on the order returned from api-server
 	addresses := make([]string, count)
 	for i := 0; i < count; i++ {
-		node := (*nodes)[i]
-		if len(node.Status.Addresses) == 0 || node.Status.Addresses[0].Address == "" {
-			c.config.Log.Info("could not get address from node", "node", node.Name)
+		nodeAddresses := (*nodes)[i]
+		if len(nodeAddresses) == 0 || nodeAddresses[0].Address == "" {
+			c.config.Log.Info("ignoring node without IP address")
```
slintes, line 196: I'm wondering how useful, if not important, it is to log the node name here?
If we want to log it, we could make the addresses a `map[string][]v1.NodeAddress`

slintes (commented):
overall looks good to me, but one suggestion inside

pkg/apicheck/check.go
n1r1: Not very important IMHO. It's probably not common to have node without IP. 
Also, it's not very difficult to find a node without IP manually

slintes: ok

PR #92, opened 2021-08-08
Run remediation only for nodes who have the poison pill pod

n1r1:
if the unhealthy node doesn't have the poison-pill pod we don't want to delete the node, since it might never be in a safe state (i.e. rebooted)

### DIFF
--- a/controllers/controller_test.go
+++ b/controllers/controller_test.go
@@ -2,19 +2,19 @@ package controllers_test
 
 import (
 	"context"
-	"github.com/medik8s/poison-pill/controllers"
-	"k8s.io/apimachinery/pkg/labels"
-	"k8s.io/apimachinery/pkg/selection"
 	"time"
 
 	. "github.com/onsi/ginkgo"
 	. "github.com/onsi/gomega"
 
+	"k8s.io/apimachinery/pkg/labels"
+	"k8s.io/apimachinery/pkg/selection"
 	v1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 	"sigs.k8s.io/controller-runtime/pkg/controller/controllerutil"
 
+	"github.com/medik8s/poison-pill/controllers"
 	poisonpillv1alpha1 "github.com/medik8s/poison-pill/api/v1alpha1"
 )
 
--- a/controllers/controller_test.go
+++ b/controllers/controller_test.go
@@ -3,6 +3,8 @@ package controllers_test
 import (
 	"context"
 	"github.com/medik8s/poison-pill/controllers"
+	"k8s.io/apimachinery/pkg/labels"
+	"k8s.io/apimachinery/pkg/selection"
 	"time"
 
 	. "github.com/onsi/ginkgo"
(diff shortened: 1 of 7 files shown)


controllers/controller_test.go
```diff
@@ -59,9 +95,41 @@ var _ = Describe("ppr Controller", func() {
 			Expect(node.CreationTimestamp).ToNot(BeZero())
 		})
 
+		Context("simulate daemonset pods assigned to nodes", func() {
```
slintes, line 98: this context looks a bit irritating on a first look... but I have no better idea without restructuring the complete test, so works for me...

controllers/controller_test.go
```diff
@@ -3,6 +3,8 @@ package controllers_test
 import (
 	"context"
 	"github.com/medik8s/poison-pill/controllers"
+	"k8s.io/apimachinery/pkg/labels"
```
slintes, line 6: I know this is picky, but do you mind sorting the imports? :)

controllers/poisonpillremediation_controller.go
```diff
@@ -19,6 +19,8 @@ package controllers
 import (
 	"context"
 	"errors"
+	"k8s.io/apimachinery/pkg/labels"
```
slintes, line 22: same here

controllers/poisonpillremediation_controller.go
```diff
@@ -240,6 +255,27 @@ func (r *PoisonPillRemediationReconciler) Reconcile(ctx context.Context, req ctr
...
 }
 
+func (r *PoisonPillRemediationReconciler) hasPoisonPillAgentPod(node *v1.Node) (bool, error) {
+	podList := &v1.PodList{}
+
+	selector := labels.NewSelector()
+	requirement, _ := labels.NewRequirement("app", selection.Equals, []string{"poison-pill-agent"})
+	selector = selector.Add(*requirement)
+
+	err := r.Client.List(context.Background(), podList, &client.ListOptions{LabelSelector: selector})
```
slintes, line 265: maybe add the namespace to the ListOptions to make it easier for the api server to look for the pods?

controllers/suite_test.go
```diff
@@ -19,6 +19,7 @@ package controllers_test
 import (
 	"context"
 	"errors"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
```
slintes, line 22: same

controllers/controller_test.go
n1r1: sure, I don't mind. 
they were automatically added by goland.

what order would you expect?

slintes: It should at least align to the other imports which already exist. What I saw most often is:
- go standard imports
- everything else
- k8s
- ocp
- own project

slintes: (and yes, we need a goland plugin for this ;) )

beekhof: personally i prefer sorting reverse alphabetically from right to left

controllers/poisonpillremediation_controller.go
```diff
@@ -172,6 +175,18 @@ func (r *PoisonPillRemediationReconciler) Reconcile(ctx context.Context, req ctr
...
+		return ctrl.Result{}, err
+	}
+
+	//if the unhealthy node doesn't have the poison pill agent pod, the node might not reboot, and we might end up
+	//in deleting a running node
+	if !hasPoisonPillPod {
+		r.logger.Error(errors.New("node is missing poison pill agent pod, which means the node might not reboot when we'll delete the node. Skipping remediation"), "")
+		return ctrl.Result{}, nil
+	}
+
```
beekhof, line 189: Will the error make it into the console, or do we need to raise an event for that?
Maybe record it in the status somewhere?

It's something you really want to find out about before the node dies

n1r1: > Will the error make it into the console

nope, only to the logs.

> do we need to raise an event for that?

events are short lived, unless pushed to some external destination.

> Maybe record it in the status somewhere?

Might be good to add this to the ppr status.

n1r1:
I'm going to merge this today so I can unblock @shellykatz who needs this functionality

There are two remaining issues that should be addressed in a follow up PR:

1. update PPR status with an error if fencing didn't take place due to missing pp pod
2. I'm thinking maybe we need to check that the pod is in running state, not just exists. It could be OOMKilled or something that would prevent it from running. I'm a bit worried about false positives where we think the pod is not running but it is. I should check what's the status of the pod after it loses api server connectivity. IIRC it should be `running`

PR #97, opened 2021-08-30
No software reboot mode new

ShellyKatz:
allow no-software-reboot mode.
update poison pill agent's label "is-reboot-capable" to be true if there's watchdog or software reboot wasn't disabled by the user (in the IsSoftwareRebootEnabled env). Otherwise, update the label to be false.

### DIFF
--- a/controllers/poisonpillremediation_controller.go
+++ b/controllers/poisonpillremediation_controller.go
@@ -28,9 +28,7 @@ import (
 	v1 "k8s.io/api/core/v1"
 	apiErrors "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
-	"k8s.io/apimachinery/pkg/labels"
 	"k8s.io/apimachinery/pkg/runtime"
-	"k8s.io/apimachinery/pkg/selection"
 	ctrl "sigs.k8s.io/controller-runtime"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 	"sigs.k8s.io/controller-runtime/pkg/controller/controllerutil"
@@ -256,24 +254,16 @@ func (r *PoisonPillRemediationReconciler) Reconcile(ctx context.Context, req ctr
 }
 
 func (r *PoisonPillRemediationReconciler) hasPoisonPillAgentPod(node *v1.Node) (bool, error) {
-	podList := &v1.PodList{}
-
-	selector := labels.NewSelector()
-	requirement, _ := labels.NewRequirement("app", selection.Equals, []string{"poison-pill-agent"})
-	selector = selector.Add(*requirement)
-
-	err := r.Client.List(context.Background(), podList, &client.ListOptions{LabelSelector: selector})
+	pod, err := utils.GetPoisonPillAgentPod(node.Name, r.Client)
 	if err != nil {
 		r.logger.Error(err, "failed to list poison pill agent pods")
 		return false, err
 	}
 
-	for _, pod := range podList.Items {
-		if pod.Spec.NodeName == node.Name {
-			return true, nil
-		}
+	if pod == nil {
+		return false, nil
 	}
-	return false, nil
+	return true, nil
 }
 
 //returns the lastHeartbeatTime of the first condition, if exists. Otherwise returns the zero value
--- a/controllers/poisonpillremediation_controller.go
+++ b/controllers/poisonpillremediation_controller.go
@@ -253,7 +253,7 @@ func (r *PoisonPillRemediationReconciler) Reconcile(ctx context.Context, req ctr
 	return ctrl.Result{RequeueAfter: 1 * time.Second}, nil
 }
 
-func (r *PoisonPillRemediationReconciler) hasPoisohasRunningPoisonPillAgentPodnPillAgentPod(node *v1.Node) (bool, error) {
+func (r *PoisonPillRemediationReconciler) hasRunningPoisonPillAgentPod(node *v1.Node) (bool, error) {
 	pod, err := utils.GetPoisonPillAgentPod(node.Name, r.Client)
 	if err != nil {
 		r.logger.Error(err, "failed to list poison pill agent pods")
--- a/main.go
+++ b/main.go
@@ -20,6 +20,8 @@ import (
 	"context"
 	"flag"
 	"fmt"
+	"github.com/medik8s/poison-pill/pkg/utils"
+	v1 "k8s.io/api/core/v1"
 	"os"
 	"strconv"
 	"time"
@@ -175,6 +177,7 @@ func initPoisonPillAgent(mgr manager.Manager) {
 		os.Exit(1)
 	}
 
+	watchdogInitiated := false
 	wd, err := watchdog.NewLinux(ctrl.Log.WithName("watchdog"))
 	if err != nil {
 		setupLog.Error(err, "failed to init watchdog, using soft reboot")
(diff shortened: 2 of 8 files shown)


main.go
```diff
@@ -185,6 +188,12 @@ func initPoisonPillAgent(mgr manager.Manager) {
 			os.Exit(1)
 		}
 	}
+	watchdogInitiated = true
```
slintes, line 191: I think this needs to go into the `if` block above, otherwise it's always `true`

main.go
```diff
@@ -329,3 +338,39 @@ func getDeploymentNamespace() (string, error) {
...
+
+	softwareRebootEnabledEnv := os.Getenv("SOFTWARE_REBOOT_ENABLED")
+	softwareRebootEnabled, err := strconv.ParseBool(softwareRebootEnabledEnv)
+	if err != nil {
+		setupLog.Error(err, "failed to convert env variable SOFTWARE_REBOOT_ENABLED value: %s to boolean",
+			softwareRebootEnabledEnv)
+		return err
+	}
+	if watchdogInitiated || softwareRebootEnabled {
+		err = updateLabel("is-reboot-capable", "true", pod, client)
```
slintes, line 360: please introduce a const for "is-reboot-capable"... can be re-used in tests as well

slintes: maybe also use a real boolean here, and convert to string in `updateLabel`?

slintes (changes requested):
In general loogs good to me, some comments inline. Maybe more of the new function can go the new `pods.go`?
I would also really like to see at least one simple test. Maybe you can add one to the e2e tests, just confirming the new label is set at all on a PP agent pod? The `findPPPod` might be useful there :)

config/manager/kustomization.yaml
```diff
@@ -13,4 +13,4 @@ kind: Kustomization
 images:
 - name: controller
   newName: quay.io/medik8s/poison-pill-operator
-  newTag: 0.1.3
+  newTag: 0.1.2-49-g0215e29
```
n1r1, line 16: could you please revert this change?

config/rbac/role_binding.yaml
```diff
@@ -10,3 +10,4 @@ subjects:
 - kind: ServiceAccount
   name: controller-manager
   namespace: system
+
```
n1r1, line 13: stray whitespace

controllers/poisonpillremediation_controller.go
```diff
@@ -5,7 +5,7 @@ Licensed under the Apache License, Version 2.0 (the "License");
 you may not use this file except in compliance with the License.
 You may obtain a copy of the License at
 
-    http://www.apache.org/licenses/LICENSE-2.0
+ http://www.apache.org/licenses/LICENSE-2.0
```
n1r1, line 8: can we revert this?

controllers/poisonpillremediation_controller.go
```diff
@@ -181,7 +179,16 @@ func (r *PoisonPillRemediationReconciler) Reconcile(ctx context.Context, req ctr
...
+	var hasPoisonPillPod bool
+
+	pod, err := utils.GetPoisonPillAgentPod(node.Name, r.Client)
+
+	if err != nil {
+		r.logger.Error(err, "failed to get poison pill agent pod resource")
+		hasPoisonPillPod = false
+	}
+
+	hasPoisonPillPod, err = r.hasRunningPoisonPillAgentPod(pod)
```
n1r1, line 191: if we couldn't get the pod previously it doesn't make sense to check if it's running.
We should `return ctrl.Result{}, err` if we got an error from `GetPoisonPillAgentPod()`

controllers/poisonpillremediation_controller.go
```diff
@@ -193,6 +200,14 @@ func (r *PoisonPillRemediationReconciler) Reconcile(ctx context.Context, req ctr
 		return ctrl.Result{}, nil
 	}
 
+	//if the unhealthy node has the poison pill agent pod, but the is-reboot-capable label is unknown/false/doesn't exist
+	//the node might not reboot, and we might end up
+	//in deleting a running node
```
n1r1, line 205: this can be added to previous line

controllers/poisonpillremediation_controller.go
```diff
@@ -193,6 +200,14 @@ func (r *PoisonPillRemediationReconciler) Reconcile(ctx context.Context, req ctr
...
+	if pod.Labels == nil || pod.Labels[isRebootCapableLabel] != "true" {
+		r.logger.Error(errors.New("agent pod's isRebootCapable label is not true, which means the node might not reboot when we'll delete the node. Skipping remediation"), "")
```
n1r1, line 207: s/agent/poison pill 
might be helpful to log the value of the label, if exists

controllers/poisonpillremediation_controller.go
```diff
@@ -262,25 +277,11 @@ func (r *PoisonPillRemediationReconciler) Reconcile(ctx context.Context, req ctr
...
-	if err != nil {
-		r.logger.Error(err, "failed to list poison pill agent pods")
-		return false, err
-	}
-
-	for _, pod := range podList.Items {
-		if pod.Spec.NodeName == node.Name {
-			return pod.Status.Phase == v1.PodRunning, nil
-		}
+func (r *PoisonPillRemediationReconciler) hasRunningPoisonPillAgentPod(pod *v1.Pod) (bool, error) {
```
n1r1, line 280: this function now doesn't know about poison pill, it just gets a pod and check if it's running.
I'd rename it to `isPodRunning` or maybe remove this function and do this check in reconcile function

main.go
```diff
@@ -191,16 +193,26 @@ func initPoisonPillAgent(mgr manager.Manager) {
 		os.Exit(1)
 	}
 
+	watchdogInitiated := false
 	wd, err := watchdog.NewLinux(ctrl.Log.WithName("watchdog"))
 	if err != nil {
 		setupLog.Error(err, "failed to init watchdog, using soft reboot")
+
```
n1r1, line 200: stray whitespace

main.go
```diff
@@ -191,16 +193,26 @@ func initPoisonPillAgent(mgr manager.Manager) {
 		os.Exit(1)
 	}
 
+	watchdogInitiated := false
```
n1r1, line 196: we usually want boolean var to start with conditional verb, in this case I'd rename to `wasWatchdogInitated` or `hasValidWatchdog`

pkg/utils/labels.go
```diff
@@ -0,0 +1,52 @@
...
+	}
+	pod.Labels[labelKey] = strconv.FormatBool(labelValue)
+	//update in yaml
+	err := c.Update(context.Background(), pod)
+	return err
+}
+
+// UpdatePodIsRebootCapableLabel updates the is-reboot-capable label to be true if any kind
+// of reboot is enabled and false if there isn't watchdog and software reboot is disabled
+//TODO - change the function to only get manager instead of client and reader
```
n1r1, line 29: sounds like something we want to do now :)

pkg/utils/labels.go
```diff
@@ -0,0 +1,52 @@
...
+func UpdatePodIsRebootCapableLabel(watchdogInitiated bool, nodeName string, client client.Client, setupLog logr.Logger, reader client.Reader) error {
```
n1r1, line 30: I think it might be better to omit the `setupLog` param and just return an err, let the caller handle the logging

pkg/utils/labels.go
```diff
@@ -0,0 +1,52 @@
...
+	//get pod in order to update it's label
```
n1r1, line 31: s/it's/its

n1r1 (changes requested):
good job!
left mainly cosmetics notes.
in a follow PR please add tests that verifies the update label functionality works as expected

thanks!

controllers/poisonpillremediation_controller.go
ShellyKatz: do we still want to log the error before returning? (line 187) or delete that and just  <return ctrl.Result{}, err>?

n1r1: still log

PR #106, opened 2021-10-04
Include release-* branches when pushing tags

rgolangh:
tags are pushed to release branches - this change should make sure github publishes official version containers on it.


rgolangh:
@n1r1 @slintes I guess the only way to test it is to merge and then push a dummy tag to see if we are happy with the result

slintes:
> the only way to test it is to merge...

and to backport to a release branch 😉 

> and then push a dummy tag

/lgtm

Issue #103, opened 2021-09-24
[RFE] Handle VolumeAttachments explicitly to speed up failover


dyasny:
Based on the discussion at https://groups.google.com/g/medik8s/c/_esi9rj9mTU/m/_aGvInpSAgAJ

When a node suddenly dies, the pods using RWO volumes on it cannot be rescheduled, because of the multi-attach error. The VolumeAttachment object effectively locks the pod to the node and until it is removed, the volume cannot be rescheduled. This generally makes sense when the RWO volumes are local, but much less so for network-attached RWO volumes.

In any case, the flow currently is for the SBA mechanism to run it's course, removing the node from the cluster, which in turn eventually removes volumeattachments (or not, depending on settings and random gods of k8s API interconnections). This bring the potential downtime of a pod from a possible few seconds, to at least 5 minutes and more (the rescheduling pod retries at an exponentially growing retry timeouts as the reconciliation loop is set). 

This RFE is about deleting the VolumeAttachments belonging to the removed node explicitly to avoid the huge delays and shorten the downtime of a workload.

n1r1:
Thanks for bringing this up @dyasny 
We will explore `volumeattachements` and `pods` direct deletion in the near future as an alternative for `node` deletion

n1r1:
closing this as it was implemented in https://github.com/medik8s/poison-pill/pull/110

PR #110, opened 2021-11-09
Add new remediation strategy - pod and volumeattachments deletion

n1r1:
This is addressing https://github.com/medik8s/poison-pill/issues/103

This PR adds a new remediation strategy to the PPR/PPRT, allowing the user to decide which remediation strategy will be used. By default, we're still going with Node deletion.

The new mode - ResourcesDeletion, has some advantages comparing to node deletion:

1. it should be faster to recover workloads as we directly remove the pods/vas, instead of deleting the node and waiting for some other controller to realize that the pods on that node were deleted.
2. it's less likely to create issues with other 3rd parties, which doesn't take into account the possibility of node being deleted and restored. For example, this will allow OVN users to run poison pill.

Tests will be added in a follow up PR, to avoid making this one too big

### DIFF
--- a/bundle/manifests/poison-pill.medik8s.io_poisonpillremediations.yaml
+++ b/bundle/manifests/poison-pill.medik8s.io_poisonpillremediations.yaml
@@ -37,6 +37,18 @@ spec:
             type: object
           spec:
             description: PoisonPillRemediationSpec defines the desired state of PoisonPillRemediation
+            properties:
+              remediationStrategy:
+                default: NodeDeletion
+                description: RemediationStrategy is the remediation method for unhealthy
+                  nodes could be either "NodeDeletion" or "ResourcesDeletion" the
+                  first will delete the node to signal to the cluster that the node
+                  was fenced the latter will iterate over all pos and volumeattachments
+                  related to the unhealthy node and delete them
+                enum:
+                - NodeDeletion
+                - ResourcesDeletion
+                type: string
             type: object
           status:
             description: PoisonPillRemediationStatus defines the observed state of
--- a/bundle/manifests/poison-pill.medik8s.io_poisonpillremediations.yaml
+++ b/bundle/manifests/poison-pill.medik8s.io_poisonpillremediations.yaml
@@ -41,13 +41,13 @@ spec:
               remediationStrategy:
                 default: NodeDeletion
                 description: RemediationStrategy is the remediation method for unhealthy
-                  nodes could be either "NodeDeletion" or "ResourcesDeletion" the
-                  first will delete the node to signal to the cluster that the node
-                  was fenced the latter will iterate over all pos and volumeattachments
+                  nodes could be either "NodeDeletion" or "ResourceDeletion" the first
+                  will delete the node to signal to the cluster that the node was
+                  fenced the latter will iterate over all pos and volumeattachments
                   related to the unhealthy node and delete them
                 enum:
                 - NodeDeletion
-                - ResourcesDeletion
+                - ResourceDeletion
                 type: string
             type: object
           status:
--- a/bundle/manifests/poison-pill.medik8s.io_poisonpillremediationtemplates.yaml
+++ b/bundle/manifests/poison-pill.medik8s.io_poisonpillremediationtemplates.yaml
@@ -45,6 +45,19 @@ spec:
                   spec:
                     description: PoisonPillRemediationSpec defines the desired state
                       of PoisonPillRemediation
+                    properties:
+                      remediationStrategy:
+                        default: NodeDeletion
+                        description: RemediationStrategy is the remediation method
+                          for unhealthy nodes could be either "NodeDeletion" or "ResourcesDeletion"
+                          the first will delete the node to signal to the cluster
+                          that the node was fenced the latter will iterate over all
+                          pos and volumeattachments related to the unhealthy node
+                          and delete them
+                        enum:
+                        - NodeDeletion
+                        - ResourcesDeletion
+                        type: string
                     type: object
                 required:
                 - spec
--- a/bundle/manifests/poison-pill.medik8s.io_poisonpillremediationtemplates.yaml
+++ b/bundle/manifests/poison-pill.medik8s.io_poisonpillremediationtemplates.yaml
@@ -49,14 +49,14 @@ spec:
                       remediationStrategy:
                         default: NodeDeletion
                         description: RemediationStrategy is the remediation method
-                          for unhealthy nodes could be either "NodeDeletion" or "ResourcesDeletion"
+                          for unhealthy nodes could be either "NodeDeletion" or "ResourceDeletion"
                           the first will delete the node to signal to the cluster
                           that the node was fenced the latter will iterate over all
                           pos and volumeattachments related to the unhealthy node
                           and delete them
                         enum:
                         - NodeDeletion
-                        - ResourcesDeletion
+                        - ResourceDeletion
                         type: string
                     type: object
                 required:
--- a/controllers/poisonpillremediation_controller.go
+++ b/controllers/poisonpillremediation_controller.go
@@ -219,17 +219,17 @@ func (r *PoisonPillRemediationReconciler) remediateWithResourcesDeletion(ppr *v1
 		}
 	}
 
-	volumeattachments := storagev1.VolumeAttachmentList{}
-	if err := r.Client.List(context.Background(), &volumeattachments); err != nil {
-		r.logger.Error(err, "failed to get volumeattachments list")
+	volumeAttachments := &storagev1.VolumeAttachmentList{}
+	if err := r.Client.List(context.Background(), volumeAttachments); err != nil {
+		r.logger.Error(err, "failed to get volumeAttachments list")
 		return ctrl.Result{}, err
 	}
 
-	for _, va := range volumeattachments.Items {
+	for _, va := range volumeAttachments.Items {
 		if va.Spec.NodeName == node.Name {
 			err = r.Client.Delete(context.Background(), &va)
 			if err != nil {
-				r.logger.Error(err, "failed to delete volumeattachment", "name", va.Name)
+				r.logger.Error(err, "failed to delete volumeAttachment", "name", va.Name)
 				return ctrl.Result{}, err
 			}
 		}
@@ -356,7 +356,7 @@ func (r *PoisonPillRemediationReconciler) remediateWithNodeDeletion(ppr *v1alpha
 	return ctrl.Result{RequeueAfter: 1 * time.Second}, nil
 }
 
-//wasNodeRebooted returns true if the node assumed to been rebooted.
+// wasNodeRebooted returns true if the node assumed to been rebooted.
 // if not, it will also return the remaining time for that to happen
 func (r *PoisonPillRemediationReconciler) wasNodeRebooted(ppr *v1alpha1.PoisonPillRemediation) (bool, time.Duration) {
 	maxNodeRebootTime := ppr.Status.TimeAssumedRebooted
--- a/controllers/poisonpillremediation_controller.go
+++ b/controllers/poisonpillremediation_controller.go
@@ -159,7 +159,8 @@ func (r *PoisonPillRemediationReconciler) remediateWithResourcesDeletion(ppr *v1
 	}
 
 	if !r.isNodeRebootCapable(node) {
-		return ctrl.Result{RequeueAfter: 10 * time.Second}, nil
+		//use err to trigger exponential backoff
+		return ctrl.Result{}, errors.New("Node is not capable to reboot itself")
 	}
 
 	if !controllerutil.ContainsFinalizer(ppr, PPRFinalizer) {
@@ -313,7 +314,7 @@ func (r *PoisonPillRemediationReconciler) remediateWithNodeDeletion(ppr *v1alpha
 	}
 
 	if !r.isNodeRebootCapable(node) {
-		return ctrl.Result{RequeueAfter: 10 * time.Second}, nil
+		return ctrl.Result{}, errors.New("Node is not capable to reboot itself")
 	}
 
 	if !controllerutil.ContainsFinalizer(ppr, PPRFinalizer) {
--- a/controllers/poisonpillremediation_controller.go
+++ b/controllers/poisonpillremediation_controller.go
@@ -127,10 +127,10 @@ func (r *PoisonPillRemediationReconciler) Reconcile(ctx context.Context, req ctr
 	r.mutex.Unlock()
 
 	switch ppr.Spec.RemediationStrategy {
-	case v1alpha1.NodeDeletionRemdiationStrategy:
+	case v1alpha1.NodeDeletionRemediationStrategy:
 		return r.remediateWithNodeDeletion(ppr)
-	case v1alpha1.ResourcesDeletionRemediationStrategy:
-		return r.remediateWithResourcesDeletion(ppr)
+	case v1alpha1.ResourceDeletionRemediationStrategy:
+		return r.remediateWithResourceDeletion(ppr)
 	default:
 		r.logger.Info("Encountered unsupported remediation strategy. Please check template spec", "strategy", ppr.Spec.RemediationStrategy)
 	}
@@ -142,7 +142,7 @@ func (r *PoisonPillRemediationReconciler) isPprCompleted(ppr *v1alpha1.PoisonPil
 	return ppr.Status.Phase != nil && *ppr.Status.Phase == completedPhase
 }
 
-func (r *PoisonPillRemediationReconciler) remediateWithResourcesDeletion(ppr *v1alpha1.PoisonPillRemediation) (ctrl.Result, error) {
+func (r *PoisonPillRemediationReconciler) remediateWithResourceDeletion(ppr *v1alpha1.PoisonPillRemediation) (ctrl.Result, error) {
 	if r.isPprCompleted(ppr) {
 		if controllerutil.ContainsFinalizer(ppr, PPRFinalizer) {
 			if err := r.removeFinalizer(ppr); err != nil {
--- a/controllers/poisonpillremediation_controller.go
+++ b/controllers/poisonpillremediation_controller.go
@@ -189,7 +189,7 @@ func (r *PoisonPillRemediationReconciler) remediateWithResourceDeletion(ppr *v1a
 
 	if r.MyNodeName == node.Name {
 		// we have a problem on this node, reboot!
-		return ctrl.Result{}, r.Rebooter.Reboot()
+		return r.rebootIfNeeded(ppr)
 	}
 
 	wasRebooted, timeLeft := r.wasNodeRebooted(ppr)
@@ -331,7 +331,7 @@ func (r *PoisonPillRemediationReconciler) remediateWithNodeDeletion(ppr *v1alpha
 
 	if r.MyNodeName == node.Name {
 		// we have a problem on this node, reboot!
-		return ctrl.Result{}, r.Rebooter.Reboot()
+		return r.rebootIfNeeded(ppr)
 	}
 
 	wasRebooted, timeLeft := r.wasNodeRebooted(ppr)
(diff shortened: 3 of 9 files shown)


bundle/manifests/poison-pill.medik8s.io_poisonpillremediations.yaml
```diff
@@ -37,6 +37,18 @@ spec:
...
+              remediationStrategy:
+                default: NodeDeletion
+                description: RemediationStrategy is the remediation method for unhealthy
+                  nodes could be either "NodeDeletion" or "ResourcesDeletion" the
+                  first will delete the node to signal to the cluster that the node
+                  was fenced the latter will iterate over all pos and volumeattachments
+                  related to the unhealthy node and delete them
+                enum:
+                - NodeDeletion
+                - ResourcesDeletion
```
beekhof, line 50: Singular for consistency

```suggestion
                - ResourceDeletion
```

bundle/manifests/poison-pill.medik8s.io_poisonpillremediationtemplates.yaml
```diff
@@ -45,6 +45,19 @@ spec:
...
+                        default: NodeDeletion
+                        description: RemediationStrategy is the remediation method
+                          for unhealthy nodes could be either "NodeDeletion" or "ResourcesDeletion"
+                          the first will delete the node to signal to the cluster
+                          that the node was fenced the latter will iterate over all
+                          pos and volumeattachments related to the unhealthy node
+                          and delete them
+                        enum:
+                        - NodeDeletion
+                        - ResourcesDeletion
```
beekhof, line 59:
```suggestion
                        - ResourceDeletion
```

controllers/poisonpillremediation_controller.go
```diff
@@ -122,6 +126,144 @@ func (r *PoisonPillRemediationReconciler) Reconcile(ctx context.Context, req ctr
...
+
+	if !node.Spec.Unschedulable || !utils.TaintExists(node.Spec.Taints, NodeUnschedulableTaint) {
+		return r.markNodeAsUnschedulable(node)
+	}
+
+	if ppr.Status.TimeAssumedRebooted.IsZero() {
+		//todo this also updates the node but we don't need it
+		return r.updatePprStatus(node, ppr)
+	}
+
```
beekhof, line 180: Feels like some/all/much of this could be common to both code paths and better located in `Reconcile()`. No?

controllers/poisonpillremediation_controller.go
```diff
@@ -122,6 +126,144 @@ func (r *PoisonPillRemediationReconciler) Reconcile(ctx context.Context, req ctr
...
+	if !node.Spec.Unschedulable || !utils.TaintExists(node.Spec.Taints, NodeUnschedulableTaint) {
+		return r.markNodeAsUnschedulable(node)
+	}
+
+	if ppr.Status.TimeAssumedRebooted.IsZero() {
+		//todo this also updates the node but we don't need it
+		return r.updatePprStatus(node, ppr)
+	}
```
beekhof, line 179: Should this happen before you mark it as unschedulable?

controllers/poisonpillremediation_controller.go
```diff
@@ -122,6 +126,144 @@ func (r *PoisonPillRemediationReconciler) Reconcile(ctx context.Context, req ctr
...
+	completed := completedPhase
+	ppr.Status.Phase = &completed
+	if err := r.Client.Status().Update(context.Background(), ppr); err != nil {
+		if apiErrors.IsConflict(err) {
+			// conflicts are expected since all poison pill deamonset pods are competing on the same requests
+			return ctrl.Result{RequeueAfter: 1 * time.Second}, nil
+		}
+		r.logger.Error(err, "failed to mark PPR as completed")
+		return ctrl.Result{}, err
+	}
```
beekhof, line 261: What is the risk if this fails?  Will it not look like the node still needs to be fenced at line 172 since the node is no longer unschedulable?

`if !node.Spec.Unschedulable || !utils.TaintExists(node.Spec.Taints, NodeUnschedulableTaint) {`

bundle/manifests/poison-pill.medik8s.io_poisonpillremediations.yaml
n1r1: I was wondering about that.
we delete a single node, but multiple resources (pods and volume).
would you still think that the singular form is better?

controllers/poisonpillremediation_controller.go
n1r1: that's a good point but I don't think it possible without over-engineering this.
I thought to create an interface for remediation strategy and to implement the strategy-specific code there, and the reconcile will have the common flow and just call the interface methods (e.g. pre-remediation, fence, post-remediation)

However, if you look closely you'll see that there are subtle differences between the two, and no one guarantees that a future remediation strategy will need the same steps, so I preferred to extract all common code into functions, and duplicate most of the function calls.
In addition, if we'll have a common code for both, we would need to consider how each code change affects both which might be more error prone.

Here's an example for a difference between the two:
NodeDeletion cares about node being deleted, as it's a valid part of the flow:
```
	node, err := r.getNodeFromPpr(ppr)
	if err != nil {
		if apiErrors.IsNotFound(err) {
			//as part of the remediation flow, we delete the node, and then we need to restore it
			return r.handleDeletedNode(ppr)
		}
		r.logger.Error(err, "failed to get node", "node name", ppr.Name)
		return ctrl.Result{}, err
	}
```

However, in ResourcesDeletion, node is not expected to be deleted at all, so it would be an error:

```
	node, err := r.getNodeFromPpr(ppr)
	if err != nil {
		r.logger.Error(err, "failed to get node", "node name", ppr.Name)
		return ctrl.Result{}, err
	}
```

controllers/poisonpillremediation_controller.go
n1r1: That's what we did so far in node-deletion.
It's a bit safer to do it after marking it unschedulable.

controllers/poisonpillremediation_controller.go
n1r1: yeah, I think you're right and this is a bug that needs to be fixed.
in node deletion we don't have this part.
anyway, we need to distinguish between a node that was fenced and is now running and a node that wasn't fenced yet.
In node deletion we use the node's creationTimestamp to detect that this is a new node.
Not sure how to do that here.

controllers/poisonpillremediation_controller.go
```diff
@@ -122,6 +126,144 @@ func (r *PoisonPillRemediationReconciler) Reconcile(ctx context.Context, req ctr
 	lastSeenPprNamespace = req.Namespace
 	r.mutex.Unlock()
 
+	switch ppr.Spec.RemediationStrategy {
+	case "NodeDeletion":
```
slintes, line 130: I think these values should be constants in the api package

controllers/poisonpillremediation_controller.go
```diff
@@ -178,73 +315,31 @@ func (r *PoisonPillRemediationReconciler) Reconcile(ctx context.Context, req ctr
...
-	if node.Annotations == nil || node.Annotations[utils.IsRebootCapableAnnotation] != "true" {
-		annVal := ""
-		if node.Annotations != nil {
-			annVal = node.Annotations[utils.IsRebootCapableAnnotation]
-		}
-		r.logger.Error(errors.New("node's isRebootCapable annotation is not `true`, which means the node might not reboot when we'll delete the node. Skipping remediation"),
-			"", "annotation value", annVal)
-		return ctrl.Result{}, nil
+	if !r.isNodeRebootCapable(node) {
+		return ctrl.Result{RequeueAfter: 10 * time.Second}, nil
```
slintes, line 320: in the old version we don't requeue, is the change on purpose?

controllers/poisonpillremediation_controller.go
```diff
@@ -178,73 +315,31 @@ func (r *PoisonPillRemediationReconciler) Reconcile(ctx context.Context, req ctr
...
-		r.logger.Info("waiting for unschedulable taint to appear", "node name", node.Name)
-		return ctrl.Result{RequeueAfter: 1 * time.Second}, nil
-	}
-
 	if ppr.Status.NodeBackup == nil || ppr.Status.TimeAssumedRebooted.IsZero() {
 		return r.updatePprStatus(node, ppr)
 	}
 
-	maxNodeRebootTime := ppr.Status.TimeAssumedRebooted
+	if r.MyNodeName == node.Name {
```
slintes, line 335: Oh, interesting, in the old version we waited for TimeAssumedRebooted for no reason, right?

controllers/poisonpillremediation_controller.go
```diff
@@ -265,6 +360,76 @@ func (r *PoisonPillRemediationReconciler) Reconcile(ctx context.Context, req ctr
 	return ctrl.Result{RequeueAfter: 1 * time.Second}, nil
 }
 
+//wasNodeRebooted returns true if the node assumed to been rebooted.
+// if not, it will also return the remaining time for that to happen
```
slintes, line 364: nit: I understand there are no rules for having a space after `//` or not, but mixing it looks strange. Personnaly I prefer with space, but I'm fine without.

slintes (commented):
1st review on the old strategy only, will review the new one in the afternoon, running out of time out :)

controllers/poisonpillremediation_controller.go
n1r1: definitely. thanks

controllers/poisonpillremediation_controller.go
n1r1: no, we didn't wait :)
we rebooted instantly. it's a bit confusing.

n1r1: the difference is that previously we would reboot only if the time to assume that node has been rebooted didn't pass.
Now we will reboot as long as the PPR still exists (and poison pill is running on the unhealthy node)

controllers/poisonpillremediation_controller.go
n1r1: The change is on purpose and mainly comes from the code refactoring, not from a real need.
do you see any problem with this?

controllers/poisonpillremediation_controller.go
```diff
@@ -122,6 +126,144 @@ func (r *PoisonPillRemediationReconciler) Reconcile(ctx context.Context, req ctr
...
+	backgroundDeletePolicy := metav1.DeletePropagationBackground
+
+	deleteOptions := &client.DeleteAllOfOptions{
+		ListOptions: client.ListOptions{
+			FieldSelector: fields.SelectorFromSet(fields.Set{"spec.nodeName": node.Name}),
+			Namespace:     "",
+			Limit:         0,
+		},
+		DeleteOptions: client.DeleteOptions{
+			GracePeriodSeconds: &zero,
```
slintes, line 204: maybe a bit nicer:
```
import "k8s.io/utils/pointer"
...
GracePeriodSeconds: pointer.Int64Ptr(0),
```

controllers/poisonpillremediation_controller.go
```diff
@@ -122,6 +126,144 @@ func (r *PoisonPillRemediationReconciler) Reconcile(ctx context.Context, req ctr
...
+	for _, ns := range namespaces.Items {
+		deleteOptions.Namespace = ns.Name
+		err = r.Client.DeleteAllOf(context.Background(), pod, deleteOptions)
+		if err != nil {
+			r.logger.Error(err, "failed to delete pods of unhealthy node", "namespace", ns.Name)
+			return ctrl.Result{}, err
+		}
+	}
+
+	volumeattachments := storagev1.VolumeAttachmentList{}
```
slintes, line 225: - uppercase A in `volumeAttachments`
- for consistency maybe make this a pointer, like with the pod above?

controllers/poisonpillremediation_controller.go
slintes: > Not sure how to do that here

Introduce a new phase `node fenced`, only mark the node schedulable when that phase was successfully set, and check that phase in line 172?
(on a side note: conditions are preferred over phase, but that's another story :) )

slintes (commented):
some nits, but otherwise the code looks good to me so far

controllers/poisonpillremediation_controller.go
slintes: ah yes, I misread the time check 😎

controllers/poisonpillremediation_controller.go
```diff
@@ -122,6 +126,144 @@ func (r *PoisonPillRemediationReconciler) Reconcile(ctx context.Context, req ctr
...
+			if err := r.removeFinalizer(ppr); err != nil {
+				return ctrl.Result{}, err
+			}
+		}
+	}
+
+	if r.isPprCompleted(ppr) {
+		return ctrl.Result{}, nil
+	}
+
```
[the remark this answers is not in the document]
n1r1: sure, thanks :)

controllers/poisonpillremediation_controller.go
```diff
@@ -367,7 +532,16 @@ func (r *PoisonPillRemediationReconciler) getNodeFromMachine(ref metav1.OwnerRef
 	return node, nil
 }
 
+//the unhealthy node might reboot itself and take new workloads
+//since we're going to delete the node eventually, we must make sure the node is deleted only
+//when there's no running workload there. Hence we mark it as unschedulable.
+//markNodeAsUnschedulable sets node.Spec.Unschedulable which triggers node controller to add the taint
 func (r *PoisonPillRemediationReconciler) markNodeAsUnschedulable(node *v1.Node) (ctrl.Result, error) {
+	if node.Spec.Unschedulable {
+		r.logger.Info("waiting for unschedulable taint to appear", "node name", node.Name)
```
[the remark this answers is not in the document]
n1r1: yeah I agree...
not sure how to resolve this without over complicating stuff or double checking

controllers/poisonpillremediation_controller.go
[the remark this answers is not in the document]
n1r1: fixed

controllers/poisonpillremediation_controller.go
n1r1: fixed

controllers/poisonpillremediation_controller.go
```diff
@@ -18,14 +18,16 @@ package controllers
 
 import (
 	"context"
-	"errors"
+	"k8s.io/apimachinery/pkg/fields"
```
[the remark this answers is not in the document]
n1r1: fixed

controllers/poisonpillremediation_controller.go
n1r1: fixed

controllers/poisonpillremediation_controller.go
n1r1: fixed

controllers/poisonpillremediation_controller.go
n1r1: yes, I'm not sure which way is better, I'll revert to the previous implementation where we just ignore it

n1r1: just to make sure we're aligned - the "endless loop" is a reconcile once in 10 seconds.

n1r1: I think I'll return an error to use exponential backoff.

slintes: > the "endless loop" is a reconcile once in 10 seconds until PPR gets deleted

yes

>  I'll return an error to use exponential backoff

👍🏼

slintes:
Code wise this looks good to me now, but I didn't test it.
Talking about tests: do you plan to add unit or e2e tests for this?

n1r1:
> Talking about tests: do you plan to add unit or e2e tests for this?

absolutely. see my first comment in this PR, specifically:

> Tests will be added in a follow up PR, to avoid making this one too big

FWIW I manually tested this for now.

slintes:
> > Talking about tests: do you plan to add unit or e2e tests for this?
> 
> absolutely. see my first comment in this PR, specifically:
> 
> > Tests will be added in a follow up PR, to avoid making this one too big
> 
> FWIW I manually tested this for now.

Ah, sorry, missed that.
/lgtm

Because of the `ResourcesDeletion` vs `ResourceDeletion` discussion, which seems unresolved, I'll put a hold on it. (I have no strong preference here, I would probably take `Resource`, but just because it's shorter 😉  ) 
/hold

bundle/manifests/poison-pill.medik8s.io_poisonpillremediations.yaml
n1r1: fixed

controllers/poisonpillremediation_controller.go
slintes: oh, I almost missed this one, I think it should be addressed?

n1r1: yep

n1r1: @slintes I believe 5147a1bb106ebf33955bc8bd4500ae84508d9299 fixes this, could you please review?

slintes: I think that commit is shifting the issue to another place now: when the status update fails, the node can be rebooted again in line 192 by a fresh PP agent pod on the affected node? We also delete VAs and pods again, but that isn't an issue IMHO.

Thinking this further: don't we always have a race here, also when the status update works? Quick reboot and new PP pod starting, versus slow VA / pod deletion and status update?

n1r1: new PP pod will be placed on the unhealthy node only after we removed the unschedulable taint, so I think we don't have an issue here, as we remove it AFTER the ppr status was updated.

slintes: are you sure?
"The default scheduler ignores unschedulable Nodes when scheduling daemon set pods."
https://docs.openshift.com/container-platform/4.9/nodes/jobs/nodes-pods-daemonsets.html

n1r1: > "The default scheduler ignores unschedulable Nodes when scheduling daemon set pods."
I find this statement very confusing :) if a scheduler ignores unschedulable Nodes when scheduling, it sounds like it won't use these nodes when scheduling.

anyway, it looks like you're right. I tried to mark a node as unshcuedlable and delete its poison pill pod, and I saw it was recreated immediately. 

we always assume this is not the case, and I wonder why we didn't see multiple reboots for the same host so far :thinking:

n1r1: just to clarify - this potentially impacts also the node deletion scenario we use today.
Maybe this can be avoided by checking the uptime of the machine. if uptime > ppr.CreationTimestamp, no reboot is required

slintes: > I find this statement very confusing :) 

ha, now when you say it, I agree 🙂 

> this potentially impacts also the node deletion scenario

I think the `node.CreationTimestamp.After(ppr.CreationTimestamp.Time)` check prevents it there

n1r1: indeed, but the poison pill pod could come up even before the node was deleted.
we wait ~3 minutes. if the node has completed reboot by then, it could reboot itself again

n1r1: @slintes 
I've added uptime check in 04816127a83cae7a28c9face58b1e1c2e80968c6.
WDYT?

pkg/utils/uptime.go
```diff
@@ -0,0 +1,19 @@
+package utils
+
+import (
+	"syscall"
```
slintes, line 4: this package is deprecated, please use `golang.org/x/sys/unix`, see https://pkg.go.dev/syscall

slintes (changes requested):
Good idea with checking uptime. One comment about the used package though, otherwise lgtm

pkg/utils/uptime.go
n1r1: right. good catch. fixed

slintes (approved):
/lgtm

dyasny:
Having two strategies is nice but why make it an OR? In classic fencing we have multiple tiers, if one fails we try the next one

n1r1:
In the two modes the node is being fenced by the same way - reboot.
Remediation strategy is basically just the way to signal kubernetes that the node was fenced and it's safe to reschedule the workloads elsewhere.

What would be the benefit to delete the node after the pods and VAs already deleted?

slintes:
> multiple tiers, if one fails we try the next one

We have something like this in the backlog of NodeHealthCheck, there we could trigger PP with different strategies (which doesn't make much sense as Nir explained), or even completely different remediators

dyasny:
> What would be the benefit to delete the node after the pods and VAs already deleted?

When tackling this issue, I wrote a very basic script that monitored the status of the nodes and if one was NotReady or Down for longer than 60 seconds, it would try to delete the volumeAttachments on it, so that the workloads would be able to reschedule. Which didn't work (every time). Looking in a bit deeper, I saw that VAs get restored faster than a newly scheduled workload takes over the volume in some cases. And as the remediation loop grows longer, the harder it is for the workload to reschedule. In this case, deleting the node helped reset the state of things. 

I didn't investigate too deeply into why this happened on some setups and didn't on others (I was running with multiple k8s distros - k3s, basic kubeadm, kops, eks, aks, microk8s and multiple versions of each).

n1r1:
> it would try to delete the volumeAttachments on it

did it also try to delete the pods on the unhealthy node? if yes, which resource was deleted first? va or pod?

>  I saw that VAs get restored faster than a newly scheduled workload takes over the volume in some cases

I'm not sure how node deletion fixes this.

PR #111, opened 2021-11-10
Remove service account yaml

ShellyKatz:
removed config/rbac/service_account.yaml and bundle/manifests/poison-pill-controller-manager_v1_serviceaccount.yaml files.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -158,12 +158,12 @@ rm -rf $$TMP_DIR ;\
 endef
 
 .PHONY: bundle ## Generate bundle manifests and metadata, then validate generated files.
-bundle: manifests kustomize
-	operator-sdk generate kustomize manifests -q
+bundle: manifests operator-sdk kustomize
+	$(OPERATOR_SDK) generate kustomize manifests -q
 	cd config/manager && $(KUSTOMIZE) edit set image controller=$(IMG)
 	sed -r -i "s|createdAt: \".*\"|createdAt: \"`date "+%Y-%m-%d %T" `\"|;" ./config/manifests/bases/poison-pill.clusterserviceversion.yaml
-	$(KUSTOMIZE) build config/manifests | envsubst | operator-sdk generate bundle -q --overwrite --version $(VERSION) $(BUNDLE_METADATA_OPTS)
-	operator-sdk bundle validate ./bundle
+	$(KUSTOMIZE) build config/manifests | envsubst | $(OPERATOR_SDK) generate bundle -q --overwrite --version $(VERSION) $(BUNDLE_METADATA_OPTS)
+	$(OPERATOR_SDK) bundle validate ./bundle
 
 .PHONY: bundle-build ## Build the bundle image.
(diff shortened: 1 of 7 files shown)


slintes (changes requested):
I think removing `config/rbac/service_account.yaml` isn't a good idea, because that will break deployment via `make deploy`. I knew we don't have that manifest in the bundle of NMO, and wanted to find out why... turns out that it's just the version of `operator-sdk`, version 1.12.0 does not put the service account manifest into the bundle 🤷🏼‍♂️ 
Suggestion: use a fixed version of `operator-sdk`, which is installed by the `Makefile`. See https://github.com/kubevirt/node-maintenance-operator/blob/master/Makefile#L230 and all lines containing `OPERATOR_SDK`. Then delete the bundle dir, run `make bundle`, and see if the result is good 🙂

n1r1:
> I think removing `config/rbac/service_account.yaml` isn't a good idea, because that will break deployment via `make deploy`. I knew we don't have that manifest in the bundle of NMO, and wanted to find out why... turns out that it's just the version of `operator-sdk`, version 1.12.0 does not put the service account manifest into the bundle 🤷🏼‍♂️ Suggestion: use a fixed version of `operator-sdk`, which is installed by the `Makefile`. See https://github.com/kubevirt/node-maintenance-operator/blob/master/Makefile#L230 and all lines containing `OPERATOR_SDK`. Then delete the bundle dir, run `make bundle`, and see if the result is good slightly_smiling_face


CVP will start failing operators with service account that is already defined in CSV, so this is a must, at least downstream.
and forward looking, this will be a must in OSDK and having that SA file is against their recommendation.
for more details see:
https://issues.redhat.com/browse/ECOPROJECT-352
https://github.com/operator-framework/operator-sdk/issues/5244

I didn't think about the `make deploy` case TBH.
I prefer to install the bundle. maybe it worth taking this to the operator-sdk team

n1r1:
@ShellyKatz thanks for the PR!
I believe there are two non-related commits here that should be removed: 
c5db7e4 and  e89bc4c which describe metrics stuff

slintes:
It has to be removed from `/bundle`, that's what OLM / CVP etc sees, because OLM will create the service account for you.
But it should not be removed from `/config`, that will break dev flow using `make deploy`, which does not use OLM, so we don't get a service account for free.

n1r1:
> It has to be removed from `/bundle`, that's what OLM / CVP etc sees, because OLM will create the service account for you. But it should not be removed from `/config`, that will break dev flow using `make deploy`, which does not use OLM, so we don't get a service account for free.

correct. but we need to make sure that `make bundle` doesn't copy it again from the config folder
cc @ShellyKatz

slintes:
yes, and that is just updating operator-sdk...

slintes:
@ShellyKatz automatic backport to the release-0.3 branch doesn't work, can you create a new PR targeting that branch manually please?

PR #112, opened 2021-11-10
Edit csv

ShellyKatz:
added "created at" field (the date is the date the build was moved to QE) and added PP to the keywords


slintes (changes requested):
This change will be overridden by the next run of `make bundle`. You need to mofify `config/manifests/bases/poison-pill.clusterserviceversion.yaml`.
And the date should be added dynamically instead of hard coded IMHO... I have no quick idea what's the best way to do this though 🤔

ShellyKatz:
> This change will be overridden by the next run of `make bundle`. You need to mofify `config/manifests/bases/poison-pill.clusterserviceversion.yaml`. And the date should be added dynamically instead of hard coded IMHO... I have no quick idea what's the best way to do this though thinking

PR #113, opened 2021-11-18
update CreatedAt in CSV and and add "PP" as a keyword

ShellyKatz:
This PR adds "PP" to the keywords in the CSV files (for easier search in OperatorHub).
It also adds the "createdAt" field to the CSV files that update automatically on "make bundle".

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -161,6 +161,7 @@ endef
 bundle: manifests kustomize
 	operator-sdk generate kustomize manifests -q
 	cd config/manager && $(KUSTOMIZE) edit set image controller=$(IMG)
+	sed -r -i "s|createdAt: \".*\"|createdAt: \"`date "+%Y-%m-%d %T" `\"|;" ./config/manifests/bases/poison-pill.clusterserviceversion.yaml
 	$(KUSTOMIZE) build config/manifests | envsubst | operator-sdk generate bundle -q --overwrite --version $(VERSION) $(BUNDLE_METADATA_OPTS)
 	operator-sdk bundle validate ./bundle
 
--- a/bundle/manifests/poison-pill.clusterserviceversion.yaml
+++ b/bundle/manifests/poison-pill.clusterserviceversion.yaml
@@ -37,11 +37,12 @@ metadata:
     capabilities: Basic Install
     categories: OpenShift Optional
     containerImage: quay.io/medik8s/poison-pill-operator:0.1.3
+    createdAt: "2021-11-22 14:21:44"
     operatorframework.io/suggested-namespace: poison-pill
+    operators.openshift.io/infrastructure-features: '["disconnected"]'
     operators.operatorframework.io/builder: operator-sdk-v1.9.0+git
     operators.operatorframework.io/project_layout: go.kubebuilder.io/v3
     repository: https://github.com/medik8s/poison-pill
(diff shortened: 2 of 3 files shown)


Makefile
```diff
@@ -161,6 +161,8 @@ endef
 bundle: manifests kustomize
 	operator-sdk generate kustomize manifests -q
 	cd config/manager && $(KUSTOMIZE) edit set image controller=$(IMG)
+	sed -r -i "s|createdAt: \".*\"|createdAt: \"`date "+%Y-%m-%d %T" `\"|;" ./config/manifests/bases/poison-pill.clusterserviceversion.yaml
+	sed -r -i "s|createdAt: \".*\"|createdAt: \"`date "+%Y-%m-%d %T" `\"|;" ./bundle/manifests/poison-pill.clusterserviceversion.yaml
```
slintes, line 165: I think you can skip this line, the following line will copy the manifest from `/config` to `/bundle`, including the up-to-date `createdAt`

Makefile
```diff
@@ -161,6 +161,8 @@ endef
 bundle: manifests kustomize
 	operator-sdk generate kustomize manifests -q
 	cd config/manager && $(KUSTOMIZE) edit set image controller=$(IMG)
+	sed -r -i "s|createdAt: \".*\"|createdAt: \"`date "+%Y-%m-%d %T" `\"|;" ./config/manifests/bases/poison-pill.clusterserviceversion.yaml
```
slintes, line 164: 👍🏼

slintes (changes requested):
In general this is a nice solution 👍🏼 Small nit inside.
I'm only wondering how we can make sure, especially downstream, that we have a bundle with a up-to-date `createdAt` date, because IIRC downstream we don't run `make bundle`? @n1r1, WDYT?

n1r1:
> In general this is a nice solution 👍🏼 Small nit inside. I'm only wondering how we can make sure, especially downstream, that we have a bundle with a up-to-date `createdAt` date, because IIRC downstream we don't run `make bundle`? @n1r1, WDYT?

I usually run diff between upstream and downstream CSV and take only the changes I'm interested in, so I think this should be okay.

n1r1:
/approve
@slintes please add lgtm when you think this is okay

n1r1:
/hold
why this PR has 10 commits? looks like there are non-related stuff there.

ShellyKatz:
@n1r1 I think it's because when I tried to push my last commit it said that my local branch head is behind the remote one (not sure why..) and probably while fixing it, it created the same commits again.
Can I just delete the commits that repeat themselves? which are:
1) added created at field and added PP to the keywords
0efa02f
2) added PP to keyword in config/manifests/bases/poison-pill.clusterserv… 
fb7e3fa
3) Added createdAt field to csv in config/manifests/bases. Added sed com… 
9d74545

n1r1:
@ShellyKatz let's solve this together later today, if that works for you

slintes:
> > I'm only wondering how we can make sure, especially downstream, that we have a bundle with a up-to-date `createdAt` date, because IIRC downstream we don't run `make bundle`? @n1r1, WDYT?
> 
> I usually run diff between upstream and downstream CSV and take only the changes I'm interested in, so I think this should be okay.

yeah, annoying manual process, but I don't have a better idea
lgtm (not as a command because of the needed cleanup)

PR #114, opened 2021-11-18
added the annotation to support disconnected mode

ShellyKatz:
This PR adds a label that enables the operator to work in a disconnected environment


ShellyKatz:
@n1r1  I added it manually to the config/manifests/bases/poison-pill.clusterserviceversion.yaml

n1r1:
@ShellyKatz and then it was copied to the other file after `make bundle`?

slintes:
> @slintes I guess you're doing similar stuff for NMO, can you have a look at this?

Looks good to me, and yes we need this for NMO (@razo7) and NHC as well, thanks for the heads up!

> @ShellyKatz and then it was copied to the other file after `make bundle`?

I'm not Shelly, but yes 😉  `make bundle` fills the `/bundle` dir based on what we have in `/config`

/lgtm

PR #116, opened 2021-12-05
Add Resource Deletion tests

n1r1:
Refactored existing tests and added tests for the recently added mode - pod and va deletion

### DIFF
--- a/controllers/controller_test.go
+++ b/controllers/controller_test.go
@@ -323,24 +323,18 @@ func verifyPoisonPillPodDoesntExist() {
 func verifyNodeIsSchedulable() {
 	By("Verify that node is not marked as unschedulable")
 	node := &v1.Node{}
-	Eventually(func() bool {
+	Eventually(func() (bool, error) {
 		err := k8sClient.Client.Get(context.TODO(), unhealthyNodeNamespacedName, node)
-		if err != nil {
-			return true
-		}
-		return node.Spec.Unschedulable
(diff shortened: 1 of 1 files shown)


controllers/controller_test.go
```diff
@@ -266,14 +274,160 @@ var _ = Describe("ppr Controller", func() {
...
+	}, 5*time.Second, 250*time.Millisecond).Should(BeTrue())
+}
+
+func verifyNodeIsSchedulable() {
+	By("Verify that node is not marked as unschedulable")
+	node := &v1.Node{}
+	Eventually(func() bool {
+		err := k8sClient.Client.Get(context.TODO(), unhealthyNodeNamespacedName, node)
+		if err != nil {
+			return true
```
slintes, line 329: just fyi: I recently learned that the functions for `Eventually` can return more then one value, which helps preventing "workarounds" like this for dealing with errors: `If multple values are returned by the function, Eventually will pass the first value to the matcher and require that all others are zero-valued. This allows you to pass Eventually a function that returns a value and an error - a common pattern in Go.`
Applies to some other places as well.

slintes (approved):
/lgtm
one information inside, I put a hold in case you want to address it in this PR

controllers/controller_test.go
n1r1: changed in 29864ec81aba7e4e218041259276f821107eb842

PR #122, opened 2021-12-28
Fix potential double unlock

mshitrit:
In order to prevent a potential unknown behavior occurring on a second mutex unlock  (triggered by the `defer` block), the potential second unlock is triggered after verifying the mutex is locked.


mshitrit:
After some discussion:
Since this use case only expected to occur in case another panic has already happened, I'm not certain that the risk of introducing new code overcomes the risk of the problem it should solve- closing this PR

PR #126, opened 2022-01-02
Return and log error for (invalid) wrong remediation strategy use case

mshitrit:
We enforce valid options for remediation strategy by kubebuilder, so there shouldn't be a use-case where we get an invalid one.
In case we still do, logging and returning an error (instead of current info message).

### DIFF
--- a/controllers/poisonpillremediation_controller.go
+++ b/controllers/poisonpillremediation_controller.go
@@ -132,10 +132,12 @@ func (r *PoisonPillRemediationReconciler) Reconcile(ctx context.Context, req ctr
 	case v1alpha1.ResourceDeletionRemediationStrategy:
 		return r.remediateWithResourceDeletion(ppr)
 	default:
-		r.logger.Info("Encountered unsupported remediation strategy. Please check template spec", "strategy", ppr.Spec.RemediationStrategy)
+		//this should never happen since we enforce valid values with kubebuilder
+		err := errors.New("unsupported remediation strategy")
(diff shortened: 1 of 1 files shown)


controllers/poisonpillremediation_controller.go
```diff
@@ -132,10 +132,12 @@ func (r *PoisonPillRemediationReconciler) Reconcile(ctx context.Context, req ctr
 	case v1alpha1.ResourceDeletionRemediationStrategy:
 		return r.remediateWithResourceDeletion(ppr)
 	default:
-		r.logger.Info("Encountered unsupported remediation strategy. Please check template spec", "strategy", ppr.Spec.RemediationStrategy)
+		//this should never happen since we enforce valid values with kubebuilder
+		err := errors.New("unsupported remediation strategy")
+		r.logger.Error(err, "Encountered unsupported remediation strategy. Please check template spec", "strategy", ppr.Spec.RemediationStrategy)
+		return ctrl.Result{}, err
```
n1r1, line 138: is there any reason to requeue?
Both strategy and code aren't going to change

mshitrit: Since this is an error that shouldn't occur I think in that case it's ok to give it more visibility in the unlikely chance that it does occur.

n1r1: Okay. Up to you

slintes: How does this give more visibility? 🤔 
IMHO more visibilty would be a) Updating some status or b) throwing an event.
But since in this case the issue should not occur anyway, not worth more thoughts probably.
Just in general I agree with Nir: don't requeue when nothing can make the issue to disappear from alone.

PR #129, opened 2022-02-01
Remove service account yaml

ShellyKatz:
updated operator-sdk version to 1.12.0 in makefile and then ran rm bundle dir and recreated it with: make bundle VERSION=0.1.3


ShellyKatz:
made a mess with the commits.. ignore that please

PR #130, opened 2022-02-01
remove service account YAML from bundle 

ShellyKatz:
updated operator-sdk version to 1.12.0 in makefile and then ran rm bu…ndle dir and recreated it with: make bundle VERSION=0.1.3


slintes:
release branch was re-created, not needed anymore :) 

/close

PR #132, opened 2022-02-03
Update poison pill owners

mshitrit:
Updating PP owners


slintes (changes requested):
Why the 2nd commit? If there is a good reason for it, It should be a PR on its own.

PR #145, opened 2022-02-16
remove suggestion namespace

razo7:
Remove suggestion namespace, since Poison Pill is using AllNamespaces installMode.

Signed-off-by: oraz <oraz@redhat.com>

PR #146, opened 2022-02-16
add support annotation

razo7:
Signed-off-by: oraz <oraz@redhat.com>


mshitrit:
/hold

I assume this change is part of aligning medik8s components right ?
I've noticed I didn't see this change in the master, is it intentionally pushed just to the release ?

razo7:
> I assume this change is part of aligning medik8s components right ?

Yes and no. Our components are missing this annotation, thus it is for the current release (although PP has it in midstream), and also for aligning our team's products.

>  I've noticed I didn't see this change in the master, is it intentionally pushed just to the release ?
Yes, I have done it as well for #145 and #146 because I though that after we release PP v0.3.0 we will backport these changes to `master`. Should I do it differently?

PR #144, opened 2022-02-16
Remove-manager_auth_proxy_patch

razo7:
Remove redundant rbac proxy container

PR #149, opened 2022-02-17
remove rbac container proxy

razo7:
- Moving manager's container ports and env variables from `manager_auth_proxy_patch.yaml` to new files `manager_ports_patch.yaml`, `manager_env_patch.yaml`.
- Disable `manager_auth_proxy_patch.yaml` which removes redundant rbac proxy container (fix problem in #144 PR).

Signed-off-by: oraz <oraz@redhat.com>

### DIFF
--- a/bundle/manifests/poison-pill.clusterserviceversion.yaml
+++ b/bundle/manifests/poison-pill.clusterserviceversion.yaml
@@ -286,17 +286,6 @@ spec:
             spec:
               containers:
               - args:
-                - --secure-listen-address=0.0.0.0:8443
-                - --upstream=http://127.0.0.1:8080/
-                - --logtostderr=true
-                - --v=10
-                image: gcr.io/kubebuilder/kube-rbac-proxy:v0.8.0
-                name: kube-rbac-proxy
-                ports:
-                - containerPort: 8443
-                  name: https
-                resources: {}
-              - args:
                 - --health-probe-bind-address=:8081
                 - --metrics-bind-address=127.0.0.1:8080
                 - --leader-elect
--- a/config/default/kustomization.yaml
+++ b/config/default/kustomization.yaml
@@ -28,7 +28,7 @@ patchesStrategicMerge:
 # Protect the /metrics endpoint by putting it behind auth.
 # If you want your controller-manager to expose the /metrics
 # endpoint w/o any authn/z, please comment the following line.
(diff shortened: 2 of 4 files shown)


slintes:
/hold

And again: look at the diff. It's removing more than just the proxy container!

bundle/manifests/poison-pill.clusterserviceversion.yaml
```diff
@@ -314,30 +314,10 @@ spec:
...
-                - --is-manager
                 command:
                 - /manager
                 env:
-                - name: MY_NODE_NAME
-                  valueFrom:
-                    fieldRef:
-                      fieldPath: spec.nodeName
-                - name: POISON_PILL_IMAGE
-                  value: quay.io/medik8s/poison-pill-operator:0.1.3
```
beekhof, line 340: This isn't part of the proxy though. Did you mean to delete it?

razo7: No, that was a mistake.

config/default/manager_auth_proxy_patch.yaml
```diff
@@ -9,16 +9,6 @@ spec:
   template:
     spec:
       containers:
-      - name: kube-rbac-proxy
```
slintes, line 12: The name of this patch file is manager_auth_proxy_patch, and you remove exactly that now ;)
Please make a new patch file with all changes NOT related to the proxy, disabled this one, and enable the new patch file. The history of this file should show what needs to be in the new patch. It's not your fault, this was done wrong in the past already (adding things not related to the proxy container to this patch). Let's make it right at least now, so we don't run into confusion again in case we want the proxy container back in future.

slintes (changes requested):

config/default/manager_auth_proxy_patch.yaml
razo7: I have updated that and created two files instead of one, following NMO's motivation in `kustomization.yaml` https://github.com/medik8s/node-maintenance-operator/blob/master/config/default/kustomization.yaml#L31

slintes (commented):
lgtm
/hold cancel

leaving the official lgtm to @mshitrit

slintes (approved):

slintes:
that wasn't planned this way (removing the earlier "request changes" = lgtm) 😄

PR #151, opened 2022-02-21
[WIP] Split daemonset & manager SA

mshitrit:
At the moment the same SA is used both by the operator and the DS.
The purpose of this PR is to split them and make sure each has the necessary permissions.

- as a first step I'm attempting just to split up the SA while each still have all the permissions (both SA are still connected to the same role).

### DIFF
--- a/bundle/manifests/poison-pill.clusterserviceversion.yaml
+++ b/bundle/manifests/poison-pill.clusterserviceversion.yaml
@@ -37,7 +37,7 @@ metadata:
     capabilities: Basic Install
     categories: OpenShift Optional
     containerImage: quay.io/medik8s/poison-pill-operator:0.1.3
-    createdAt: "2022-02-15 17:15:21"
+    createdAt: "2022-03-27 10:14:53"
     olm.skipRange: '>=0.1.0 <0.3.0'
     operators.openshift.io/infrastructure-features: '["disconnected"]'
     operators.operatorframework.io/builder: operator-sdk-v1.16.0+git
@@ -170,6 +170,12 @@ spec:
           - update
           - watch
         - apiGroups:
+          - ""
+          resources:
+          - serviceaccounts
+          verbs:
+          - create
+        - apiGroups:
           - machine.openshift.io
           resources:
           - machines
@@ -267,6 +273,18 @@ spec:
           - get
           - patch
           - update
+        - apiGroups:
+          - rbac.authorization.k8s.io
+          resources:
+          - clusterrolebindings
+          verbs:
+          - create
+        - apiGroups:
+          - rbac.authorization.k8s.io
+          resources:
+          - clusterroles
+          verbs:
+          - '*'
         - apiGroups:
           - security.openshift.io
           resourceNames:
--- a/config/manifests/bases/poison-pill.clusterserviceversion.yaml
+++ b/config/manifests/bases/poison-pill.clusterserviceversion.yaml
@@ -6,7 +6,7 @@ metadata:
     capabilities: Basic Install
     categories: OpenShift Optional
     containerImage: quay.io/medik8s/poison-pill-operator:0.0.0
-    createdAt: "2022-02-15 17:15:21"
+    createdAt: "2022-03-27 11:50:42"
     olm.skipRange: '>=0.1.0 <0.3.0'
     operators.openshift.io/infrastructure-features: '["disconnected"]'
     repository: https://github.com/medik8s/poison-pill
--- a/config/rbac/role.yaml
+++ b/config/rbac/role.yaml
@@ -60,6 +60,12 @@ rules:
   - update
   - watch
 - apiGroups:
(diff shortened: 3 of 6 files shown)


bundle/manifests/poison-pill-daemonset_v1_serviceaccount.yaml
```diff
@@ -0,0 +1,5 @@
+apiVersion: v1
+kind: ServiceAccount
+metadata:
+  creationTimestamp: null
+  name: poison-pill-daemonset
```
beekhof, line 5: You're missing the file in `config/` that this is generated from

bundle/manifests/poison-pill-daemonset-rolebinding_rbac.authorization.k8s.io_v1_clusterrolebinding.yaml
```diff
@@ -0,0 +1,13 @@
...
+  creationTimestamp: null
+  name: poison-pill-daemonset-rolebinding
+roleRef:
+  apiGroup: rbac.authorization.k8s.io
+  kind: ClusterRole
+  name: poison-pill-manager-role
+subjects:
+- kind: ServiceAccount
+  name: poison-pill-daemonset
+  namespace: poison-pill
```
beekhof, line 13: [config/rbac/daemonset_role_binding.yaml](https://github.com/medik8s/poison-pill/pull/151/files#diff-c43e51c7d58d5b61956a52a9b35611353e14f2f258fbcbd8d705b471e040d615) says `namespace: system`

Which one is correct?

bundle/manifests/poison-pill-daemonset_v1_serviceaccount.yaml
beekhof: And it would need to be added to `config/rbac/kustomization.yaml`

mshitrit: Hmm.
 IIUC [here](https://github.com/medik8s/poison-pill/pull/151/files#diff-331a7d0d6ec2e3237b4cfec828db02c693512ea311da9c88f8d0fb8a2d146f66) is the file that this is generated from and [here](https://github.com/medik8s/poison-pill/pull/151/files#diff-3fd064f2f1027a954b035a6d5885b96cc5b483f04889d14fa840ace5e9c29c55R13) it was added to kustomization.yaml. 
What am I missing ? :thinking:

bundle/manifests/poison-pill-daemonset-rolebinding_rbac.authorization.k8s.io_v1_clusterrolebinding.yaml
slintes: both: system is overwritten by this: https://github.com/medik8s/poison-pill/blob/master/config/default/kustomization.yaml#L2

slintes: however: this is an issue nonetheless, because AFAIK now OLM will always create this resource in the poison-pill namespace, no matter where PP is actually installed. maybe removing the namespace from the manifest solves this. Otherwise we need to create the new resources in code 🤷🏼‍♂️

slintes: looking at the e2e test, it looks like even with the poison-pill namespace something went wrong, I don't see any daemonset pod 🤔

bundle/manifests/poison-pill-daemonset_v1_serviceaccount.yaml
beekhof: I think I'm going senile :) 
Not sure what I thought I was looking at earlier but I see now its ok

bundle/manifests/poison-pill-daemonset-rolebinding_rbac.authorization.k8s.io_v1_clusterrolebinding.yaml
```diff
@@ -0,0 +1,13 @@
...
+  creationTimestamp: null
+  name: poison-pill-daemonset-rolebinding
+roleRef:
+  apiGroup: rbac.authorization.k8s.io
+  kind: ClusterRole
+  name: poison-pill-manager-role
+subjects:
+- kind: ServiceAccount
+  name: poison-pill-daemonset
+  namespace: system
```
slintes, line 13: Ah crap, this will never work I'm afraid
While the SA might be installed into the right ns without the namespace in its metadata, you need the correct namespace at runtime here, and AFAIK OLM doesn't help you with that.
I guess only solution is to create the rolebinding in code.

mshitrit:
PP not maintained

PR #152, opened 2022-02-24
Refactor render pkg

mshitrit:
- removed unused consts
- removed unused funcs
- modified public func names to comply with package name
- changed var declaration to comply with go lint


pkg/render/render.go
```diff
@@ -78,15 +71,15 @@ func RenderDir(manifestDir string, d *RenderData) ([]*unstructured.Unstructured,
...
 // json file representing one or more k8s api objects
-func RenderTemplate(path string, d *RenderData) ([]*unstructured.Unstructured, error) {
+func Template(path string, d *Data) ([]*unstructured.Unstructured, error) {
 	rendered, err := renderTemplate(path, d)
 	if err != nil {
 		return nil, err
 	}
 
-	out := []*unstructured.Unstructured{}
+	var out []*unstructured.Unstructured
```
razo7, line 82: Now we don't initialize this variable but looking at where we used to call `RenderTemplate` then it seems fine

mshitrit: It is best practice to use `nil` over an empty slice.
If you use golint you'll see a notification for that.
In case you want more info, you can check [this link](https://gosamples.dev/empty-vs-nil-slice/)

PR #157, opened 2022-03-31
WIP: Webhook for timeouts validation newest

beekhof:


bundle/manifests/poison-pill-webhook-service_v1_service.yaml
```diff
@@ -0,0 +1,13 @@
...
+kind: Service
+metadata:
+  creationTimestamp: null
+  name: poison-pill-webhook-service
+spec:
+  ports:
+  - port: 443
+    targetPort: 9443
+  selector:
+    control-plane: controller-manager
```
beekhof, line 11: I think we're missing a selector here.  See https://github.com/medik8s/node-maintenance-operator/blob/2a9f1a0ebf54f7c9962cea91ae8de809f7f3ddb8/bundle/manifests/node-maintenance-operator-webhook-service_v1_service.yaml#L15 and https://github.com/medik8s/node-maintenance-operator/blob/2a9f1a0ebf54f7c9962cea91ae8de809f7f3ddb8/config/default/kustomization.yaml#L12-L13

beekhof: That would cause requests to randomly go to nodes where the PP manager isn't running and could explain the behaviour you're seeing

config/webhook/service.yaml
```diff
@@ -0,0 +1,12 @@
...
+kind: Service
+metadata:
+  name: webhook-service
+  namespace: system
+spec:
+  ports:
+    - port: 443
+      targetPort: 9443
+  selector:
+    control-plane: controller-manager
```
beekhof, line 12: See https://github.com/medik8s/poison-pill/pull/157/files#r839149176

bundle/manifests/poison-pill-webhook-service_v1_service.yaml
slintes: heads up, adding (or modifying) labels to deployments breaks upgrades, because OLM tries to update the existing deployment, but label updates are forbidden. Workaround (AFAIK, needs to be tested): also modify the deployment name. That will make OLM to create a new deployment and delete the old one on upgrades.

beekhof: perfect timing then, since this will only go into the renamed operator

shellykatz: > also modify the deployment name
in which file do I modify and to what name should I change? @slintes

razo7: @shellykatz  I think it is here, https://github.com/medik8s/self-node-remediation/blob/main/config/manager/manager.yaml#L11 (then you need to run `make build` to apply it in other files).

shellykatz: Thanks! and what the new name should be?

PR #158, opened 2022-04-26
Setting version to 0.3.1

mshitrit:
This PR is the upstream preparation for the 0.3.1 downstream Poison Pill Version which will limit Poison Pill to OCP 4.10 or older.


razo7:
Did you run `VERSION=0.3.1 make bundle`? It will update the operator's version to `0.3.1` in more places in the CSV.

slintes:
> > Did you run `VERSION=0.3.1 make bundle`? It will update the operator's version to `0.3.1` in more places in the CSV.
> 
> IIUC we avoid pushing specific version upstream - ASAIS we didn't do it in the previous versions as well.

for whatever reason it is v0.1.3, it is not v0.3.0 indeed 🤷🏼‍♂️ 😁 

/lgtm

PR #161, opened 2022-07-28
Poison Pill deprecation notice.

mshitrit:
Adding  deprecation notice in poison pill readme.

### DIFF
--- a/README.md
+++ b/README.md
@@ -3,7 +3,7 @@
 **July 28, 2022**
 
 **Poison Pill Operator** is deprecated in favor of [Self Node Remediation](https://github.com/medik8s/self-node-remediation),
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -1,3 +1,10 @@
+# Deprecation Notice
+
+**July 28, 2022**
+
+**Poison Pill Operator** is deprecated in favor of [Self Node Remediation](https://github.com/medik8s/self-node-remediation),
+thus Poison Pill  is moving to a maintenance mode only and we are not taking PRs except for urgent bug fixes.
```
razo7, line 6: NIT: There are two spaces between `Pill` and `is` instead of one.

razo7: Do we have a support life cycle policy for PP? Meaning that we won't support PP from date x/y/z

slintes: this is upstream, no support at all 😉

